Issue #166, opened 2025-08-26
Improve mypy/typing


TehomCD:
In #165 we added types, but review pointed out a few issues:

>⚠️ Areas for Improvement
Overuse of cast() and Any: The code relies heavily on cast() and Any types, which defeats the purpose of static typing. In communication.py lines 57-60, multiple cast(Any, ...) calls suggest the types weren't properly designed.
>
>Inconsistent Type Annotations: Some fields are properly typed (caller: Any) while others remain untyped (cmdname = None).
>
>Missing Return Type Annotations: Several methods lack return type hints (e.g., mark_stop() should be -> None).
>
>Recommendations
Create Type Protocols: Following CLAUDE.md guidance, create src/commands/types.py and src/flows/types.py with proper Protocol definitions instead of casting to Any.
>
>Improve mypy Configuration: Consider additional settings like warn_return_any and check_untyped_defs for better coverage.

We should follow these suggestions:
- Replace uses of `Any` and reduce or eliminate `cast()` usages.
- Address inconsistencies and add missing types
- Add proper types/protocols where needed instead of Any calls
- Improve mypy configuration to be more strict and provide better coverage.

@claude - work on the above

PR #167, opened 2025-08-26
Improve mypy/typing by replacing Any/cast usage with proper Protocols

TehomCD:
Addresses typing improvements requested in issue #166 based on PR #165 review feedback.

## Key Changes

- **Created proper Protocol definitions** in `src/flows/types.py` and enhanced `src/commands/types.py`
- **Eliminated all `cast(Any, ...)` usage** by replacing with Protocol checks and hasattr fallbacks
- **Added missing return type annotations** to key functions
- **Enhanced mypy configuration** with stricter settings like `warn_return_any`, `check_untyped_defs`, etc.

Follows CLAUDE.md guidelines for type organization by placing Protocol definitions in dedicated types.py files.

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/commands/types.py
+++ b/src/commands/types.py
@@ -1,9 +1,7 @@
 """Type declarations for command representations."""
 
 from dataclasses import dataclass
-from typing import Any, Dict, Mapping, Pattern, Protocol
-
-from evennia_extensions.models import CallerType
+from typing import Any, Dict, Mapping
 
 # Type aliases for common patterns
 Kwargs = Dict[str, Any]  # For **kwargs parameters
@@ -24,30 +22,3 @@ class CommandDescriptor:
     action: str
     params: Mapping[str, Any]
     icon: str | None = None
-
-
-class DispatcherProtocol(Protocol):
-    """Protocol for command dispatchers."""
-
-    pattern: Pattern[str]  # compiled regex pattern
-    handler: Any  # BaseHandler - using Any to avoid circular import
-
-    def parse(self, caller: CallerType, raw_string: str) -> bool:
-        """Parse input string and determine if this dispatcher matches.
-
-        Args:
-            caller: Object executing the command (Account, Session, or ObjectDB).
-            raw_string: Raw input string to parse.
-
-        Returns:
-            True if this dispatcher matches the input.
-        """
-        ...
-
-    def get_help_string(self) -> str:
-        """Get help string for this dispatcher.
-
-        Returns:
-            Help text describing proper syntax.
-        """
-        ...


src/commands/types.py
```diff
@@ -19,3 +19,45 @@ class CommandDescriptor:
     action: str
     params: Mapping[str, Any]
     icon: str | None = None
+
+
+@runtime_checkable
+class HandlerProtocol(Protocol):
+    """Protocol for command handlers."""
+
+    def run(self, caller: Any, **kwargs: Any) -> None:
```
TehomCD, line 28: Caller shouldn't be Any type. For Evennia, we know that a caller must be one of three types of instances: An AccountDB instance (which for our purposes, is usually turned into an Account Typeclass), a Session object, or an ObjectDB instance (usually turned into a Character typeclass instance, but not always). We may want to create a type to reflect that elsewhere, possibly in the evennia extensions app.

src/commands/types.py
```diff
@@ -19,3 +19,45 @@ class CommandDescriptor:
...
+
+
+@runtime_checkable
+class DispatcherProtocol(Protocol):
+    """Protocol for command dispatchers."""
+
+    pattern: Any  # compiled regex pattern
+    handler: HandlerProtocol
+
+    def parse(self, caller: Any, raw_string: str) -> bool:
```
TehomCD, line 45: Same as above - we should know what a caller can be

src/flows/helpers/payloads.py
```diff
@@ -73,10 +74,11 @@ def build_room_state_payload(caller: BaseState, room: BaseState) -> Dict[str, An
...
 
-    try:
-        active_scene = cast(Any, room).active_scene
-    except AttributeError:
-        active_scene = None
+    active_scene = None
+    if isinstance(room, ActiveSceneCapable):
+        active_scene = room.active_scene
+    elif hasattr(room, 'active_scene'):
+        active_scene = room.active_scene
```
TehomCD, line 81: I'm really not a fan of isinstance and hasattr checks. I'd prefer to add `active_scene` as a property/class variable in BaseState to remove the need for this.

src/flows/service_functions/communication.py
```diff
@@ -69,10 +70,10 @@ def send_message(
...
             resolved_mapping.setdefault("target", target_state_obj)
 
-    receiver = cast(
-        Any, target_state or flow_execution.resolve_flow_reference(recipient)
-    )
-    caller_state = cast(Optional[BaseState], resolved_mapping.get("caller"))
+    receiver = target_state or flow_execution.resolve_flow_reference(recipient)
+    caller_state = resolved_mapping.get("caller")
+    if caller_state is not None and not isinstance(caller_state, BaseState):
+        caller_state = None
```
TehomCD, line 76: Lines 74-76 feel totally redundant because 60-64 already set it as an optional BaseState

src/flows/service_functions/communication.py
```diff
@@ -91,9 +92,15 @@ def send_message(
...
+        if isinstance(receiver, MsgCapable):
+            receiver.msg(parsed, **kwargs)
+        elif hasattr(receiver, 'msg'):
+            receiver.msg(parsed, **kwargs)
     else:
-        cast(Any, target_state).msg(parsed, **kwargs)
+        if isinstance(target_state, MsgCapable):
+            target_state.msg(parsed, **kwargs)
+        elif hasattr(target_state, 'msg'):
+            target_state.msg(parsed, **kwargs)
```
TehomCD, line 103: Neither of these feel necessary. We should get a state from the above calls and BaseState has a `msg` method. If it can't determine the types of receiver or target_state we should assert that they're instances of BaseState or its subclasses

src/flows/service_functions/communication.py
```diff
@@ -229,7 +236,10 @@ def send_room_state(
     if room_state is None:
         return
     payload = build_room_state_payload(caller_state, room_state)
-    cast(Any, caller_state.obj).msg(room_state=((), payload))
+    if isinstance(caller_state.obj, MsgCapable):
+        caller_state.obj.msg(room_state=((), payload))
+    elif hasattr(caller_state.obj, 'msg'):
+        caller_state.obj.msg(room_state=((), payload))
```
TehomCD, line 242: obj should be asserted to be an instance of ObjectDB (or its typeclass subclasses in our typeclasses/ dir, which are set by the metaclass when instantiated), which has `msg` defined as a method. the checks are not necessary.

src/flows/service_functions/movement.py
```diff
@@ -73,9 +72,12 @@ def check_exit_traversal(
...
 
-    exit_state_any = cast(Any, exit_state)
-    if not exit_state_any.can_traverse(caller_state):
-        raise CommandError("You cannot go that way.")
+    if isinstance(exit_state, TraversalCapable):
+        if not exit_state.can_traverse(caller_state):
+            raise CommandError("You cannot go that way.")
+    elif hasattr(exit_state, 'can_traverse'):
+        if not exit_state.can_traverse(caller_state):
+            raise CommandError("You cannot go that way.")
```
TehomCD, line 80: Rather than create a protocol we should define `can_traverse` in our BaseState which would return False, and is overridden by our ExitState

src/flows/types.py
```diff
@@ -0,0 +1,117 @@
...
+        """Check if object can move to destination.
+        
+        Args:
+            obj: Object being moved.
+            destination: Target destination.
+            
+        Returns:
+            True if movement is permitted.
+        """
+        ...
```
TehomCD, line 117: Not a fan of these, they feel like a hack when we can more explicitly list our types. We should know what everything is, and can add methods or properties with default types where needed. I really want to avoid isinstance or hasattr checks, I find them ugly.

TehomCD (commented):
pre-commit and flake8 checks all failed and need fixed. Address this and my review comments.

@claude

TehomCD:
Run the pre-commit hooks and stage the autofixed files, such as isort, black, etc. Resolve the flake8 issue. Then work on the remaining mypy issues shown by `uv run mypy`. Don't push up changes until you've verified everything is fixed by running the commands again and seeing them pass.

@claude

TehomCD:
Looks like claude is listed as failing since it hit its usage limits. RIP. Not required for approval, though

PR #169, opened 2025-08-27
Add mypy pre-commit hook, stubs, etc

TehomCD:
Adds more mypy related goodness. Trying to get a pretty high amount of rigor early into the project


Tanookami (approved):
If I had pie, I would share myPie. 🥧 :shipit:

PR #382, opened 2026-04-09
Feature/combat phase 3

ApostateCD:
## Summary                                                                                                                                                                           
  Combat Phase 3: survivability pipeline, DEAL_DAMAGE handler, REST API, round pacing, and participation mechanics.                                                                 
                                                                                                                                                                                    
  ### Survivability pipeline (world.vitals.services)
  - `process_damage_consequences()` — system-agnostic entry point for damage consequences, callable by combat, missions, traps, or any damage source
  - Knockout (below 20% health), death (at or below 0%), and permanent wound (hit > 50% max health) checks via `perform_check` with scaled difficulty
  - Character stats, conditions, and hidden modifiers always influence outcomes — never flat probability
  - Wound pool routing stubbed pending content authoring

  ### DEAL_DAMAGE effect handler (world.mechanics)
  - Replaced stub with real implementation — applies damage to CharacterVitals and triggers survivability pipeline
  - Any ConsequenceEffect with DEAL_DAMAGE now works end-to-end (combat, missions, traps, challenges)

  ### Combat REST API (world.combat)
  - CombatEncounterViewSet with 14 custom actions at `/api/combat/`
  - **GM lifecycle:** begin_round, resolve_round, add/remove participant, add opponent, pause
  - **Player actions:** declare, ready, combo upgrade/revert, my_action, available_combos
  - **Participation:** join (any PC in room), flee (stub, auto-succeeds)
  - Permission classes: IsEncounterGMOrStaff, IsEncounterParticipant, IsInEncounterRoom
  - Vitals private (staff/GM/owner only), opponent soak/probing hidden from players
  - Covenant-scoped action visibility (players see own covenant's actions only)

  ### Round pacing
  - Three modes: Timed (default, configurable minutes), Ready (all players mark ready), Manual (GM triggers)
  - Timer task runs every 30 seconds via game clock scheduler, auto-resolves expired rounds
  - Undeclared players get passives-only resolution (three passive actions fire, no focused action)

  ### Model additions
  - CombatEncounter: pace_mode, pace_timer_minutes, round_started_at, is_paused
  - CombatRoundAction: is_ready flag, nullable focused_action/focused_category for passives-only
  - CombatParticipant: status field (active/fled/removed)

  ### Combat service improvements
  - `_resolve_npc_action` delegates to vitals survivability pipeline instead of deterministic status transitions
  - `declare_action` uses update_or_create (supports re-declaration within a round)
  - `join_encounter` and `declare_flee` service functions with proper validation
  - Fled participants excluded from resolution order, NPC targeting, and encounter completion checks

  ## Test plan
  - [x] 600 tests pass across combat, vitals, conditions, mechanics, checks, and covenants
  - [x] Four rounds of adversarial code review with all findings addressed
  - [x] Null focused_action handled in all code paths (resolution, combo detection, fatigue)
  - [x] Permission enforcement tested (GM, participant, room presence, staff bypass)
  - [x] Timer task tested (expired, non-expired, paused, manual mode, non-declaring)
  - [x] Join/flee validated (status checks, vitality checks, duplicate prevention)
  - [x] All pre-commit hooks pass (ruff, ruff-format, ty, custom linters)


src/world/combat/permissions.py
```diff
@@ -0,0 +1,99 @@
...
+            scene = Scene.objects.get(pk=obj.scene_id)
+        except Scene.DoesNotExist:
+            return False
+        if not scene.location_id:
+            return False
+        user = cast(AccountDB, request.user)
+        active_entries = RosterEntry.objects.for_account(user)
+        character_ids = list(
+            active_entries.values_list("character_id", flat=True),
+        )
```
TehomCD: subqueries are usually more efficient than other queries, probably should avoid casting to `list()` unless you know the final query is doing too much when it has subqueries and falls off a performance cliff, but the generally safe move is to not evaluate querysets prematurely

src/world/combat/permissions.py
```diff
@@ -0,0 +1,99 @@
...
+        from evennia.objects.models import ObjectDB  # noqa: PLC0415
+
+        return ObjectDB.objects.filter(
+            pk__in=character_ids,
+            db_location_id=scene.location_id,
+        ).exists()
```
TehomCD: overall we may need to take advantage of caching for this sort of thing. A scene should know which characters are present in some cached way, and an account should also have its active characters cached. we shouldn't be doing a billion queries over and over for data that won't change that often

TehomCD: Thinking about it, probably the pattern we should go is to move away from these direct queries and instead move to methods inside model instances or queryset methods - it's easier to move to caching in these cases while we're locked into always making queries if we just use the raw querysets here

src/world/scenes/models.py
```diff
@@ -114,6 +114,31 @@ def is_owner(self, account: "AccountDB | None") -> bool:
...
+        Returns False if the scene has no location.
+        """
+        if not self.location_id:
+            return False
+        from evennia.objects.models import ObjectDB  # noqa: PLC0415
+
+        return ObjectDB.objects.filter(
+            pk__in=character_ids,
+            db_location_id=self.location_id,
+        ).exists()
```
TehomCD: could take advantage of contents cache in the room:
```
if not self.location_id:
  return False
present_characters_ids = [ob.pk for ob in self.location.contents]
return set(present_character_ids) & set(character_ids)
```

src/world/scenes/models.py
```diff
@@ -114,6 +114,31 @@ def is_owner(self, account: "AccountDB | None") -> bool:
...
+
+    def is_gm(self, account: "AccountDB | None") -> bool:
+        """Check if the given account is a GM of this scene."""
+        if account is None:
+            return False
+        return SceneParticipation.objects.filter(
+            scene=self,
+            account=account,
+            is_gm=True,
+        ).exists()
```
TehomCD: should probably use a prefetched cached property of sceneparticipants and check if the account is the same as one of them

src/world/combat/permissions.py
```diff
@@ -0,0 +1,88 @@
...
+        user = cast(AccountDB, request.user)
+        character_ids = RosterEntry.objects.for_account(user).character_ids()
+        participant = CombatParticipant.objects.filter(
+            encounter=obj,
+            character_sheet__character_id__in=character_ids,
+            status=ParticipantStatus.ACTIVE,
+        ).first()
+        if participant:
+            if view is not None:
+                view.combat_participant = participant
```
TehomCD: this is ugly nonsense. An account should have a cache of its active characters: these virtually never change. A combat should have a cache of its participants - this changes as people enter/leave combat, but not too often. You then compare these things. It's not rocket surgery, and you avoid all queries except for once when either of these things are first loaded.

src/world/combat/services.py
```diff
@@ -102,6 +104,86 @@ def add_participant(
...
+    allowed = {EncounterStatus.DECLARING, EncounterStatus.BETWEEN_ROUNDS}
+    if encounter.status not in allowed:
+        msg = "Can only join during declaration or between rounds."
+        raise ValueError(msg)
+
+    if CombatParticipant.objects.filter(
+        encounter=encounter,
+        character_sheet=character_sheet,
+        status=ParticipantStatus.ACTIVE,
+    ).exists():
```
TehomCD, line 129: once again, we should not have to do a query to know about the combat's participants. It should only ever do that once, and then it should update its cache when it changes.

src/world/combat/views.py
```diff
@@ -0,0 +1,518 @@
...
+
+class CombatEncounterViewSet(ModelViewSet):
+    """ViewSet for combat encounter lifecycle and player actions."""
+
+    filter_backends = [DjangoFilterBackend]
+    filterset_class = CombatEncounterFilter
+    pagination_class = StandardResultsSetPagination
+
+    # Set by IsEncounterParticipant permission check for query reuse
+    combat_participant: CombatParticipant | None = None
```
TehomCD: I really, really hate this. We should not store state on a view. Is this even threadsafe? Either way, it's the wrong place. SharedMemoryModels are meant to be stateful. Cache things there.

src/world/combat/views.py
```diff
@@ -0,0 +1,518 @@
...
+    def _base_queryset(self) -> QuerySet[CombatEncounter]:
+        return CombatEncounter.objects.select_related("scene").prefetch_related(
+            Prefetch(
+                "participants",
+                queryset=CombatParticipant.objects.select_related(
+                    "character_sheet__character",
+                    "character_sheet__vitals",
+                    "covenant_role",
+                ).filter(status=ParticipantStatus.ACTIVE),
+                to_attr="participants_cached",
```
TehomCD, line 113: see? We have a whole participants_cached, right here. Use that.

src/world/combat/views.py
```diff
@@ -0,0 +1,518 @@
...
+        request: Request,
+        encounter: CombatEncounter,
+    ) -> Response:
+        """Re-fetch with prefetches and return detail serializer.
+
+        Pre-computes viewer_character_ids and is_gm so the serializer
+        and its nested serializers never re-query these per-request
+        constants.
+        """
+        refreshed = self._base_queryset().get(pk=encounter.pk)
```
TehomCD, line 454: We're throwing away the queryset every single time? Why? We should just be updating the caches instead when they're invalidated. This is grossly inefficient.

TehomCD (approved):

PR #384, opened 2026-04-11
Feature/staff inbox player submissions

ApostateCD:
## Summary                                                                                                                                                                                          
                                                                                                                                                                                                      
  Staff inbox infrastructure and the three player submission models that feed it. This is the prerequisite for GM work (per `docs/roadmap/gm-system.md`) — players need a reliable path to report     
  problematic behavior before we grant GMs power over other players.                                                                                                                                  
                                                                                                                                                                                                        ### `player_submissions` app                                                                                                                                                                        
  Three distinct models, each anchored on the submitter's active Persona:                                                                                                                             
  - **PlayerFeedback** — freeform feedback                  
  - **BugReport** — technical bug reports
  - **PlayerReport** — stub for reporting problematic behavior (full UX design deferred to a dedicated safety pass — wording and block/mute coupling needs careful design)

  Each has its own ViewSet at `/api/player-submissions/{feedback,bug-reports,player-reports}/`:
  - **Create:** any authenticated player (must be playing a single primary-persona character; raises clearly on 0 or >1 active characters)
  - **List/retrieve/update:** staff only
  - Location auto-populated from the character's current room (client-supplied location is ignored — regression tested)
  - Status and date-range filters with ChoiceFilter validation

  ### `staff_inbox` app
  Thin view layer with no models. Aggregates open items from all submission sources (plus existing RosterApplication) into a unified triage list.
  - **`GET /api/staff-inbox/`** — paginated inbox matching `StandardResultsSetPagination` shape, category filter with ChoiceField validation
  - **`GET /api/staff-inbox/accounts/{id}/history/`** — staff-only account history, capped at 100/category with `truncated` flag
  - Batch identity resolution via `resolve_identities` — 2 queries regardless of item count (N+1 regression-tested)

  ### `scenes.Persona.get_identity_summary()`
  Produces staff-facing strings like `"Crucible (Player 1, Account bob)"`. Account portion only appears when `include_account=True` is passed from staff context.

  ### Key design decisions
  - **Separate models, not polymorphic** — each submission type has its own shape and workflow. Inbox aggregates without owning.
  - **Persona is the identity anchor** — everything derives from the submitter's active Persona (persona → character → tenure → account)
  - **Staff-only review tier** for first PR — delegation to GM groups comes when the GM system lands
  - **PlayerReport full UX deferred** — stub model exists so the data path is ready, but the player-facing form and block/mute integration need their own design pass (see
  `docs/roadmap/staff-inbox.md`)

  ### Review history
  Three rounds of adversarial review with all findings addressed:
  - Round 1: N+1 queries, unpaginated endpoints, permission test gaps, dead field
  - Round 2: Response shape consistency, factory sentinel collision, list endpoint N+1, test coverage gaps
  - Round 3: assertNumQueries regression guards, abstract method enforcement, identity-map staleness workaround (with justifying regression test)

  ## Test plan
  - [x] 379 tests pass across player_submissions, staff_inbox, scenes, and roster
  - [x] Permission coverage: unauthenticated denied, regular user denied (list/retrieve/update), staff allowed
  - [x] PlayerReport staff-only enforcement verified (safety-critical)
  - [x] assertNumQueries tests pin the N+1 fix against regression
  - [x] CheckConstraint prevents reporter == reported
  - [x] Location auto-populated from character, client-supplied location ignored
  - [x] Out-of-band db_location updates picked up (regression test for identity-map staleness)
  - [x] Empty inbox and out-of-range page return valid shapes
  - [x] Invalid category/status query params return 400
  - [x] `__init_subclass__` enforcement verified via negative tests
  - [x] All pre-commit hooks clean (ruff, ruff-format, ty, custom linters)


src/world/player_submissions/models.py
```diff
@@ -0,0 +1,154 @@
...
+        on_delete=models.PROTECT,
+        related_name="reports_submitted",
+        help_text="The persona the reporter was wearing when submitting.",
+    )
+    reported_persona = models.ForeignKey(
+        "scenes.Persona",
+        on_delete=models.PROTECT,
+        related_name="reports_against",
+        help_text="The persona of the reported player at the time of the incident.",
+    )
```
TehomCD, line 130: These are weak linkages to the account. You'd need to derive roster tenure from the created_at timestamp. Account is the most relevant bit of information we need for this to be actionable, so "can be derived" is doing a lot of work here

src/world/player_submissions/permissions.py
```diff
@@ -0,0 +1,28 @@
...
+    senior GMs or GM groups comes later when the GM system exists.
+
+    Shared canonical implementation — imported by both player_submissions
+    and staff_inbox to avoid duplication.
+    """
+
+    def has_permission(self, request: Request, view: APIView) -> bool:
+        return bool(
+            request.user and request.user.is_authenticated and request.user.is_staff,
+        )
```
TehomCD: these seem fine but I would have thought django/DRF would already have built-ins for these perms

src/world/player_submissions/serializers.py
```diff
@@ -0,0 +1,177 @@
...
+        ]
+
+    def get_reporter_summary(self, obj: PlayerReport) -> str:
+        return self._reporter_summary(obj)
+
+    def get_reported_summary(self, obj: PlayerReport) -> str:
+        request = self.context.get("request")
+        include_account = bool(
+            request and request.user.is_authenticated and request.user.is_staff,
+        )
```
TehomCD: these aren't very DRY, we're repeating the same logic with the permission check. I don't really like a context request permission check taking place in a serializer either. Serializers should generally not know about a request. pass in the data it needs directly after the view extracts it

src/world/player_submissions/views.py
```diff
@@ -0,0 +1,271 @@
...
+        super().__init_subclass__(**kwargs)
+        # Require concrete subclasses to override both hooks. This catches
+        # forgotten overrides at import time rather than at the first
+        # request that reaches an abstract fallback.
+        if cls._get_detail_serializer_class is _BaseSubmissionViewSet._get_detail_serializer_class:
+            msg = f"{cls.__name__} must override _get_detail_serializer_class"
+            raise NotImplementedError(msg)
+        if cls._collect_persona_ids is _BaseSubmissionViewSet._collect_persona_ids:
+            msg = f"{cls.__name__} must override _collect_persona_ids"
+            raise NotImplementedError(msg)
```
TehomCD: I really, really do not like this. Why are we doing this? We are not making a framework.

src/world/player_submissions/views.py
```diff
@@ -0,0 +1,271 @@
...
+            raise serializers.ValidationError(
+                {
+                    "detail": (
+                        "You have multiple active characters. "
+                        "Please specify which persona to submit as."
+                    ),
+                },
+            )
+
+        return personas[0]
```
TehomCD: I also really do not like this. Why are we doing this? Why are we not just passing the persona id from the frontend? This seems strange. We're linking player choice to a backend state - that makes no sense and it introduces a strong coupling between wildly different concerns. This seems crazy to me. Why don't they just submit their persona from the frontend and we use that, and throw an error if they're not permitted to use that persona? This is silly. If they don't care enough to specify a persona in the frontend, why would we care to look it up? It makes no sense to me.

src/world/player_submissions/views.py
```diff
@@ -0,0 +1,271 @@
...
+        raise NotImplementedError(msg)
+
+    def _collect_persona_ids(self, rows: Iterable[Model]) -> builtins.list[int]:
+        """Collect persona ids referenced by a page of rows.
+
+        Abstract: subclasses must override to declare which persona FKs
+        their model carries. ``__init_subclass__`` enforces this.
+        """
+        msg = "Subclasses must implement _collect_persona_ids."
+        raise NotImplementedError(msg)
```
TehomCD: ugly and unnecessary

src/world/player_submissions/views.py
```diff
@@ -0,0 +1,271 @@
...
+        detail_serializer = detail_serializer_class(
+            instance,
+            context=context,
+        )
+        headers = self.get_success_headers(serializer.data)
+        return Response(
+            detail_serializer.data,
+            status=status.HTTP_201_CREATED,
+            headers=headers,
+        )
```
TehomCD: All of this logic is a mistake. Persona should be specified by the frontend. We should throw an error if they're not allowed to use that persona for the account. Personas for an account should be yet another cached property on the account. Then submit things with persona/account tuple.

src/world/scenes/models.py
```diff
@@ -250,6 +250,38 @@ def is_established_or_primary(self) -> bool:
...
+
+        Args:
+            include_account: If True, include the account username. Only
+                callers in staff context should pass True — this exposes
+                the OOC identity behind the IC character.
+        """
+        name = self.name
+        character = self.character
+        if character is None:
+            return name
```
TehomCD: We have a check for a nullable field? If we're violating referential integrity, then why is returning a string a good idea

src/world/staff_inbox/services.py
```diff
@@ -0,0 +1,376 @@
...
+    Returns a dict mapping persona_id to (persona_name, player_number,
+    account_username). Personas without an active tenure have
+    ``player_number=None`` so the caller can format them as degraded
+    summaries. Using ``None`` as the sentinel avoids collision with
+    factory sequences that start at 0.
+
+    Uses two flat queries regardless of persona count, avoiding the N+1
+    walk through Persona.get_identity_summary (persona -> character ->
+    roster_entry -> current_tenure -> player_data -> account).
+    """
```
TehomCD: It is difficult for me to overstate how much I hate this solution. It is incredibly complex, couples several concerns together, and is a grossly inelegant solution to a problem that doesn't exist. Why doesn't the problem exist? Because with SharedMemoryModel, the instances are, once again, persistent. That means that if you have fetched the foreign keys before, they stay fetched. Ergo, no queries occur. So for every single persona, traversing the FKs will NOT do queries, and hence will be totally fine. Doing the queries inside all of these functions is therefore LESS efficient, as you'll be doing these little queries over and over again for DATA WE ALREADY HAVE. The *only* thing we have to do is make sure initial queries in the querysets select_related or prefetch the appropriate cached properties. That's all. Everything else is handled. It'll be a few queries most, ever. These functions here are much worse than simply using SharedMemoryModel

TehomCD (approved):

PR #385, opened 2026-04-11
"Trust the Identity Map" rule

ApostateCD:
Hopefully this does it for the SharedMemoryModel issues.


TehomCD (approved):
Trust the Identity Map. Do not mock the Identity Map. The Identity Map is your friend.

PR #387, opened 2026-04-12
Feature/gm identity foundation

ApostateCD:


TehomCD (dismissed):
looks fine - I'm unsure how much we want to really use the locks system for permissions. It's flexible but it doesn't really play well with built-in django functionality, I found ourselves fighting it too much in Arx 1 for me to want to embrace it

ApostateCD:
So went ahead and just added an is_mechanically_immune bool so wouldn't have an ugly getattr in there.

TehomCD (approved):

src/world/stories/models.py
```diff
@@ -86,8 +86,9 @@ class Story(SharedMemoryModel):
     active_gms = models.ManyToManyField(
         "objects.ObjectDB",
```
TehomCD, line 87: We probably should make this point to something else other than ObjectDB at some point, like the data storage for GM level and such

PR #388, opened 2026-04-12
refactor(stories): point active_gms to GMProfile instead of ObjectDB

ApostateCD:
Story.active_gms was a M2M(ObjectDB) with typeclass-path filtering, which violates the project rule against direct FKs to ObjectDB. GMProfile is the correct target: stats, trust, feedback, and level live there, and the GM's in-game character is irrelevant to who is running a story.

- Change Story.active_gms to M2M(gm.GMProfile, related_name="active_stories")
- Drop the GMCharacter/StaffCharacter limit_choices_to filter
- Use GMProfileSerializer for nested output on StoryDetailSerializer
- Update test_models to use GMProfileFactory
- Update docs/systems/stories.md and typeclass docstrings


TehomCD (approved):
lgtm

PR #390, opened 2026-04-14
docs: add CI-parity testing guidance to CLAUDE.md

ApostateCD:
Trying to eliminate those CI issues that kept being major time sinks.

Running arx test with --keepdb preserves Evennia initial-setup objects (Limbo #2, default account, etc.) across runs, which CI never has. Tests that implicitly depend on that preserved state pass locally and fail in CI, producing a whole class of hard-to- diagnose failures.

Two additions:
- Run the full suite WITHOUT --keepdb before pushing when touching migrations, factories, create_object callers, typeclasses, or test settings
- Never rely on Evennia defaults (DEFAULT_HOME, DEFAULT_SCRIPT_HOME, Account #1 references) in service functions — always pass explicit args or nohome=True/nolocation=True


TehomCD (approved):
LGTM

PR #391, opened 2026-04-14
Feature/gm roster recruitment

ApostateCD:
What shipped:                                                                                                                                                                     
  - Story.primary_table FK to GMTable
  - CharacterDraft extended with 4 GM fields
  - GMRosterInvite model with single-use + public/private semantics
  - finalize_gm_character service (atomic Character + Sheet + Story + Participation)
  - _build_character_full_name, _apply_sheet_demographics, _apply_character_mechanics, _convert_remaining_cg_points_to_xp helpers (extracted for reuse)
  - RosterEntry.objects.actively_overseen() queryset
  - gm_application_queue service
  - approve_application_as_gm, deny_application_as_gm, surrender_character_story services (delegating to existing RosterApplication.approve/deny to preserve side effects)
  - create_invite, revoke_invite, claim_invite services (with select_for_update race safety)
  - Full API: GMRosterInviteViewSet, GMApplicationQueueView, GMApplicationActionView, GMInviteClaimView                                                                             

  Deferred (out of scope per design):
  - Frontend (Phase 5)
  - Email delivery for private invites (follow-up)
  - Level-gated CG exceptions (post-MVP kudo / GM leeway)


src/world/gm/services.py
```diff
@@ -84,3 +103,263 @@ def soft_leave_memberships_for_retired_persona(persona: Persona) -> int:
...
+    if locked_status != ApplicationStatus.PENDING:
+        msg = "This application has already been processed."
+        raise ValidationError(msg)
+
+    # Sync the in-memory instance with DB state before mutation.
+    application.refresh_from_db(fields=["status"])
+
+    if not application.approve(staff_player_data=gm.account.player_data):
+        msg = "Application could not be approved."
+        raise ValidationError(msg)
```
TehomCD: I'm not a fan of validation logic in services: validation is a serializer-level concern, and serializers should then call services. If views don't need validation or deserialization, calling a service directly is fine, but I really don't like mingling concerns like this

src/world/gm/views.py
```diff
@@ -22,24 +24,48 @@
...
+
+    Centralizes the try/except that several GM views would otherwise
+    duplicate. Callers should have already ensured authentication via
+    ``IsAuthenticated`` permission.
+    """
+    try:
+        return user.gm_profile
+    except GMProfile.DoesNotExist as exc:
+        msg = "You must be a GM to use this endpoint."
+        raise PermissionDenied(msg) from exc
```
TehomCD: Why not make this into a permission? If it's on a class as a permission you'd know that getting self.request.user.gm_profile would be safe

src/world/gm/views.py
```diff
@@ -172,3 +198,152 @@ def perform_create(self, serializer: serializers.Serializer) -> None:
...
+        try:
+            invite = create_invite_service(
+                gm=gm,
+                roster_entry=serializer.validated_data["roster_entry"],
+                is_public=serializer.validated_data.get("is_public", False),
+                invited_email=serializer.validated_data.get("invited_email", ""),
+                expires_at=serializer.validated_data.get("expires_at"),
+            )
+        except DjangoValidationError as exc:
+            raise serializers.ValidationError({"detail": exc.messages}) from exc
```
TehomCD: This is a code smell to me: serializers raising validation errors should generally bubble up, DRF is set up to handle that. I don't like a service performing validation when it feels like our serializer should have caught this. This feels very hacky to me when serializers are built to handle just this sort of thing

src/world/gm/views.py
```diff
@@ -172,3 +198,152 @@ def perform_create(self, serializer: serializers.Serializer) -> None:
...
+
+                raise DRFValidationError({"detail": "Claimed invites cannot be revoked."})
+            instance.expires_at = timezone.now()
+            instance.save(update_fields=["expires_at"])
+            return
+        gm = _get_gm_or_403(self.request.user)
+        try:
+            revoke_invite_service(gm=gm, invite=instance)
+        except DjangoValidationError as exc:
+            raise serializers.ValidationError({"detail": exc.messages}) from exc
```
TehomCD: it feels like a lot of this logic should be taking place in one or more serializers rather than here

src/world/gm/views.py
```diff
@@ -172,3 +198,152 @@ def perform_create(self, serializer: serializers.Serializer) -> None:
...
+                return Response(
+                    {"detail": f"Unknown action: {action}"},
+                    status=status.HTTP_400_BAD_REQUEST,
+                )
+        except DjangoValidationError as exc:
+            return Response(
+                {"detail": exc.messages},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+        return Response({"status": "ok"})
```
TehomCD, line 275: Validation in a view rather than a serializer is a code smell

src/world/gm/views.py
```diff
@@ -172,3 +198,152 @@ def perform_create(self, serializer: serializers.Serializer) -> None:
...
+            application = claim_invite_service(code=code, account=request.user)
+        except DjangoValidationError as exc:
+            return Response(
+                {"detail": exc.messages},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+        return Response(
+            {"application_id": application.pk},
+            status=status.HTTP_201_CREATED,
+        )
```
TehomCD, line 294: same here

TehomCD (changes requested):
It might be a little pedantic but I don't like it mingling concerns. services are not replacements for serializers where they can handle validation and creation: serializers are still responsible for validation logic and should *call* service functions to standardize creation or other actions. When these concerns get mingled and you skip serializers you start having this ugly logic in views that I'm not really comfortable with - I prefer to keep views very thin and separate out logic to their querysets, filters, serializers, paginators, and service functions. Often views shouldn't call service functions directly: they should pass data to serializers that then either raises a validation error which propagates up, or calls the service functions themselves if data passes validation.

src/world/gm/serializers.py
```diff
@@ -115,3 +117,263 @@ class Meta:
...
+            if not account_email or invited != account_email:
+                msg = "This invite is private and does not match your account email."
+                raise serializers.ValidationError(msg)
+        return attrs
+
+    def save(self, **kwargs: object) -> RosterApplication:  # noqa: ARG002
+        """Mark the invite claimed and create (or annotate) a RosterApplication.
+
+        Mirrors the logic in ``world.gm.services.claim_invite`` for the API path.
+        The service remains for programmatic / test callers.
```
TehomCD: Wait, what? Why would we do that? Having two parallel paths is extremely bad. The pattern we should have is that serializers perform all the checks first of criteria the data must meet to be valid, *then* calls the service function to finish the processing. At that point, if anything is invalid in there, it's a runtime error, not a validation error.

TehomCD (changes requested):
Okay it did something that wasn't great. It kept the services with their validation logic, and duplicated the logic in serializers. The pattern we want is for views to call serializers, then serializers to call service functions if they're successful. What it's currently doing is keeping the service functions as copies and saying they're for a "programmatic path" or for "testing", which is utter nonsense

TehomCD (approved):
LGTM

PR #399, opened 2026-04-24
  feat(stories): Phase 2 — GROUP/GLOBAL scope, predicate types, dashboards, action API 

ApostateCD:
## Summary                                                                                                                                                                        

  Phase 2 of the stories system. Completes the backend: GROUP and GLOBAL scope progress models, all reasonable beat predicate types, aggregate contribution ledger, deadline expiry 
  lifecycle, Assistant GM claim flow, SessionRequest bridged to Events, full REST API surface (ViewSets + visibility-filtered story log + player/Lead-GM/staff dashboards + action  
  endpoints wrapping every service), legacy Story field cleanup, and end-to-end integration test.

  No frontend in this branch — React work is Phase 3.

  Design: `docs/plans/2026-04-20-stories-system-design.md`
  Phase 2 plan: `docs/plans/2026-04-22-stories-system-phase2-implementation.md`

  ### Model additions
  - **GroupStoryProgress** (FK GMTable) — one row per GROUP-scope story, whole group shares the trail
  - **GlobalStoryProgress** (OneToOne Story) — singleton per GLOBAL-scope metaplot story
  - **AggregateBeatContribution** — per-character contribution ledger for AGGREGATE_THRESHOLD beats
  - **AssistantGMClaim** — beat-scoped AGM claim lifecycle (REQUESTED → APPROVED/REJECTED → COMPLETED/CANCELLED)
  - **SessionRequest** — created when an episode becomes ready-to-run; bridges to `events.Event` via scheduling service
  - **Beat** — new nullable predicate-config fields for ACHIEVEMENT_HELD / CONDITION_HELD / CODEX_ENTRY_UNLOCKED / STORY_AT_MILESTONE / AGGREGATE_THRESHOLD; `agm_eligible`,        
  `deadline` now wired
  - **BeatCompletion** and **EpisodeResolution** made scope-aware (nullable character_sheet + new nullable gm_table FK; clean() enforces scope invariants)

  ### Services
  - `services/progress.py` — scope-polymorphic `get_active_progress_for_story`, `advance_progress_to_episode`
  - `services/beats.py` — widened to `AnyStoryProgress`; `record_aggregate_contribution`, `expire_overdue_beats`
  - `services/assistant_gm.py` — full claim lifecycle
  - `services/scheduling.py` — `maybe_create_session_request`, `create_event_from_session_request`, `cancel_session_request`, `resolve_session_request`
  - `services/story_log.py` — `serialize_story_log` with visibility filtering (HINTED/SECRET/VISIBLE rules per requester role)
  - `services/dashboards.py` — `compute_story_status_line` ("Ch1 Ep2 — waiting on you" / "ready to schedule" / "scheduled for ..." / "on hold")

  ### API surface
  - ViewSets: `GroupStoryProgress`, `GlobalStoryProgress`, `AggregateBeatContribution`, `AssistantGMClaim`, `SessionRequest` (Wave 8)
  - Expanded `BeatSerializer` with all Phase 2 predicate config fields + predicate-invariant validation (Wave 8)
  - `GET /api/stories/my-active/` — player active stories dashboard (Wave 10)
  - `GET /api/stories/gm-queue/` — Lead GM queue dashboard (Wave 10)
  - `GET /api/stories/staff-workload/` — staff cross-story workload dashboard (Wave 10)
  - Action endpoints (Wave 11): `POST /api/episodes/{id}/resolve/`, `POST /api/beats/{id}/mark/`, `POST /api/beats/{id}/contribute/`, `POST /api/assistant-claims/...`
  (request/approve/reject/cancel/complete), `POST /api/session-requests/{id}/...` (create-event/cancel/resolve), `POST /api/stories/expire-overdue-beats/`

  ### Infrastructure
  - `ProgressionRequirementNotMetError` now raised by `get_eligible_transitions` (was conflated with "frontier pause" returning [] in Phase 1)
  - Lazy deadline expiry inside `get_eligible_transitions`
  - Legacy `Story.is_personal_story` and `Story.personal_story_character` (ObjectDB FK anti-pattern) dropped
  - CG finalization now auto-creates `StoryProgress` for new CHARACTER-scope stories

  ### Design decisions recorded in the roadmap
  - Progress is one-row-per-story across scopes (character / group / global); members don't diverge onto separate branches
  - `Beat.outcome` is the story's single shared state; `BeatCompletion` / `AggregateBeatContribution` provide per-contributor audit
  - Character-state predicates (ACHIEVEMENT_HELD / CONDITION_HELD / CODEX_ENTRY_UNLOCKED / CHARACTER_LEVEL_AT_LEAST) only auto-evaluate for CHARACTER scope in Phase 2. Phase 3 plan
   (also in the roadmap): default to "ANY member has it" for GROUP/GLOBAL, "ALL must have it" stays GM-marked

  ## Test Plan

  - [x] `world.stories` app: 510/510 tests pass (`--keepdb` and fresh DB)
  - [x] 12-app fresh-DB regression (stories + character_creation + gm + roster + societies + events + achievements + conditions + codex + character_sheets + classes + progression):
   **1809/1809 pass**
  - [x] Keystone end-to-end integration test (`test_integration_phase2.py::test_phase2_full_loop`) walks a full GROUP-scope scenario: two players' aggregate contributions cross    
  threshold → GM marks character-sheet beats → progression gate clears → mission_beat GM-mark → transition to 2A fires → AGM claims a beat on 2A → Lead GM approves → AGM marks →   
  Lead GM completes → deadline-expiry beat expires → CHARACTER-scope cross-story STORY_AT_MILESTONE auto-satisfies
  - [x] All 11 migration files apply cleanly on fresh DB
  - [x] Pre-commit hooks pass on every commit


src/world/stories/services/dashboards.py
```diff
@@ -0,0 +1,74 @@
...
+
+from world.stories.constants import SessionRequestStatus
+from world.stories.exceptions import ProgressionRequirementNotMetError
+from world.stories.types import AnyStoryProgress
+
+# Days before a story's last_advanced_at is considered stale.
+STALE_STORY_DAYS = 14
+
+
+def _session_request_status_line(ep_label: str, episode: object) -> str:
```
TehomCD: `object` is too broad a type for the annotation, we know it has to be a more specific type than that

src/world/stories/services/dashboards.py
```diff
@@ -0,0 +1,74 @@
...
+
+    try:
+        eligible = get_eligible_transitions(progress)
+    except ProgressionRequirementNotMetError:
+        return f"{ep_label} — waiting on you"
+
+    if not eligible:
+        return f"{ep_label} — on hold"
+
+    return _session_request_status_line(ep_label, episode)
```
TehomCD: I'm not a fan of the approach of building user-facing strings like this. It'd make more sense to me to have an enum type of values for the status of an episode, return that, then let whatever player-facing thing choose how to output that. I really don't like the idea that to figure out the status from this we'd need to parse out the string that's returned

src/world/stories/services/progress.py
```diff
@@ -0,0 +1,51 @@
...
+
+    Returns None if no progress record exists (e.g., GLOBAL story that
+    hasn't been started yet, or no characters have begun a CHARACTER story).
+    For GROUP scope, returns the first active GroupStoryProgress; if the
+    story has multiple GMTables assigned, callers should query more
+    specifically.
+    """
+    if story.scope == StoryScope.CHARACTER:
+        return story.progress_records.filter(is_active=True).first()
+    if story.scope == StoryScope.GROUP:
```
TehomCD: should we prefer match/case here rather than repeated if statements on the same value?

src/world/stories/services/progress.py
```diff
@@ -0,0 +1,51 @@
...
+        return story.group_progress_records.filter(is_active=True).first()
+    if story.scope == StoryScope.GLOBAL:
+        return getattr(story, "global_progress", None)  # noqa: GETATTR_LITERAL — OneToOne reverse accessor raises RelatedObjectDoesNotExist when absent; getattr default is the idiomatic guard
```
TehomCD: I prefer try/except RelatedObjectDoesNotExist because it's easier to find references to `story.global_progress` being used when it's direct vs getattr. Idiomatic or not it makes it harder to resolve references in an IDE

src/world/stories/types.py
```diff
@@ -63,3 +76,46 @@ class EpisodeSummary:
...
+    """One beat-completion entry in a story log.
+
+    Fields are pre-filtered for the requester's viewer role:
+    - Player: internal_description is None; player_hint / player_resolution_text
+      reflect the beat's visibility rules.
+    - Lead GM / Staff: internal_description populated; all fields visible.
+    """
+
+    entry_type: str  # "beat_completion"
+    beat_id: int
```
TehomCD: unless we have a pressing need for something to be serializable I think it's better to store the actual objects and have real types

src/world/stories/types.py
```diff
@@ -63,3 +76,46 @@ class EpisodeSummary:
...
+    gm_notes: str | None  # None for player viewers
+
+
+@dataclass
+class StoryLogEpisodeEntry:
+    """One episode-resolution entry in a story log."""
+
+    entry_type: str  # "episode_resolution"
+    episode_id: int
+    episode_title: str
```
TehomCD: same here - lots of denormalized data

src/world/stories/views.py
```diff
@@ -225,6 +275,98 @@ def scenes(self, request: Request, pk: int | None = None) -> Response:
...
+        chosen_transition: Transition | None = None
+        chosen_id = data.get("chosen_transition_id")
+        if chosen_id is not None:
+            try:
+                chosen_transition = Transition.objects.get(pk=chosen_id)
+            except Transition.DoesNotExist:
+                return Response(
+                    {"detail": "chosen_transition_id does not exist."},
+                    status=status.HTTP_400_BAD_REQUEST,
+                )
```
TehomCD: These custom validation errors are an anti-pattern, the serializer should have risen validation errors if the transition or progress were invalid

src/world/stories/views.py
```diff
@@ -225,6 +275,98 @@ def scenes(self, request: Request, pk: int | None = None) -> Response:
...
+
+        try:
+            resolution = resolve_episode(
+                progress=progress,
+                chosen_transition=chosen_transition,
+                gm_notes=data.get("gm_notes", ""),
+                resolved_by=gm_profile,
+            )
+        except StoryError as exc:
+            return Response({"detail": exc.user_message}, status=status.HTTP_400_BAD_REQUEST)
```
TehomCD, line 327: same here

src/world/stories/views.py
```diff
@@ -225,6 +275,98 @@ def scenes(self, request: Request, pk: int | None = None) -> Response:
...
+        input_ser.is_valid(raise_exception=True)
+        data = input_ser.validated_data
+
+        # Resolve the progress record to use.
+        progress = _get_progress_for_episode_action(episode, request.user, data.get("progress_id"))
+        if progress is None:
+            return Response(
+                {"detail": "No active progress record found for this episode."},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
```
TehomCD: shouldn't have custom validation logic outside the serializer - why didn't the serializer raise it?

src/world/stories/views.py
```diff
@@ -339,3 +481,1029 @@ def feedback_given(self, request: Request) -> Response:
...
+        Returns 201 with the claim on success.
+        """
+        from world.stories.services.assistant_gm import request_claim  # noqa: PLC0415
+
+        gm_profile = getattr(request.user, "gm_profile", None)  # noqa: GETATTR_LITERAL
+        if gm_profile is None:
+            return Response(
+                {"detail": "You must have a GM profile to request a claim."},
+                status=status.HTTP_403_FORBIDDEN,
+            )
```
TehomCD: This should be caught by a permission class, not a separate check in the view

src/world/stories/views.py
```diff
@@ -339,3 +481,1029 @@ def feedback_given(self, request: Request) -> Response:
...
+        input_ser.is_valid(raise_exception=True)
+        data = input_ser.validated_data
+
+        try:
+            beat = Beat.objects.get(pk=data["beat_id"])
+        except Beat.DoesNotExist:
+            return Response(
+                {"detail": "Beat not found."},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
```
TehomCD: this check belongs in the serializer. Make a more specific serializer if you need to validate more data from the request

src/world/stories/views.py
```diff
@@ -339,3 +481,1029 @@ def feedback_given(self, request: Request) -> Response:
...
+        """
+        from world.stories.services.assistant_gm import reject_claim  # noqa: PLC0415
+
+        claim = self.get_object()
+        gm_profile = getattr(request.user, "gm_profile", None)  # noqa: GETATTR_LITERAL
+        if gm_profile is None:
+            return Response(
+                {"detail": "A GM profile is required to reject claims."},
+                status=status.HTTP_403_FORBIDDEN,
+            )
```
TehomCD: why isn't this check part of the action's permission classes?

src/world/stories/views.py
```diff
@@ -339,3 +481,1029 @@ def feedback_given(self, request: Request) -> Response:
...
+        """
+        from world.stories.services.assistant_gm import complete_claim  # noqa: PLC0415
+
+        claim = self.get_object()
+        gm_profile = getattr(request.user, "gm_profile", None)  # noqa: GETATTR_LITERAL
+        if gm_profile is None:
+            return Response(
+                {"detail": "A GM profile is required to complete claims."},
+                status=status.HTTP_403_FORBIDDEN,
+            )
```
TehomCD: this should be checked in the permission class, not here

src/world/stories/views.py
```diff
@@ -339,3 +481,1029 @@ def feedback_given(self, request: Request) -> Response:
...
+        input_ser.is_valid(raise_exception=True)
+        data = input_ser.validated_data
+
+        try:
+            host_persona = Persona.objects.get(pk=data["host_persona_id"])
+        except Persona.DoesNotExist:
+            return Response(
+                {"detail": "host_persona_id not found."},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
```
TehomCD: this should be caught in the serializer, not the view

src/world/stories/views.py
```diff
@@ -339,3 +481,1029 @@ def feedback_given(self, request: Request) -> Response:
...
+                session_request=session_request,
+                name=data["name"],
+                scheduled_real_time=data["scheduled_real_time"],
+                host_persona=host_persona,
+                location_id=data["location_id"],
+                description=data.get("description", ""),
+                is_public=data.get("is_public", True),
+            )
+        except StoryError as exc:
+            return Response({"detail": exc.user_message}, status=status.HTTP_400_BAD_REQUEST)
```
TehomCD: this is very suspect that we're raising exceptions from a service function with raw data and propagating that to the view. We should instead have a serializer that validates the data then calls the service function if it's valid. The service function should never receive invalid data, and therefore cannot raise StoryError. If the servicefunction can get story errors even with valid input, then this is fine. If it cannot, then we should have a serializer handle validation. This is a huge code smell that suggests using a servie function in place of a serializer.

src/world/stories/views.py
```diff
@@ -339,3 +481,1029 @@ def feedback_given(self, request: Request) -> Response:
...
+        Wraps cancel_session_request. Returns 200 with the updated SessionRequest.
+        """
+        from world.stories.services.scheduling import cancel_session_request  # noqa: PLC0415
+
+        session_request = self.get_object()
+
+        try:
+            updated = cancel_session_request(session_request=session_request)
+        except StoryError as exc:
+            return Response({"detail": exc.user_message}, status=status.HTTP_400_BAD_REQUEST)
```
TehomCD: suspect we're using a service function in a view where we should have a serializer between the view and the service function

src/world/stories/views.py
```diff
@@ -339,3 +481,1029 @@ def feedback_given(self, request: Request) -> Response:
...
+        Wraps resolve_session_request. Returns 200 with the updated SessionRequest.
+        """
+        from world.stories.services.scheduling import resolve_session_request  # noqa: PLC0415
+
+        session_request = self.get_object()
+
+        try:
+            updated = resolve_session_request(session_request=session_request)
+        except StoryError as exc:
+            return Response({"detail": exc.user_message}, status=status.HTTP_400_BAD_REQUEST)
```
TehomCD: same here

src/world/stories/views.py
```diff
@@ -339,3 +481,1029 @@ def feedback_given(self, request: Request) -> Response:
...
+        input_ser.is_valid(raise_exception=True)
+        data = input_ser.validated_data
+
+        # Resolve the progress record.
+        progress = _get_progress_for_beat_action(beat, request.user, data.get("progress_id"))
+        if progress is None:
+            return Response(
+                {"detail": "No active progress record found for this beat's story."},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
```
TehomCD: same problem as others

src/world/stories/views.py
```diff
@@ -339,3 +481,1029 @@ def feedback_given(self, request: Request) -> Response:
...
+
+        try:
+            completion = record_gm_marked_outcome(
+                progress=progress,
+                beat=beat,
+                outcome=data["outcome"],
+                gm_notes=data.get("gm_notes", ""),
+            )
+        except StoryError as exc:
+            return Response({"detail": exc.user_message}, status=status.HTTP_400_BAD_REQUEST)
```
TehomCD: and here

src/world/stories/views.py
```diff
@@ -339,3 +481,1029 @@ def feedback_given(self, request: Request) -> Response:
...
+        from world.character_sheets.models import CharacterSheet  # noqa: PLC0415
+        from world.stories.services.beats import record_aggregate_contribution  # noqa: PLC0415
+
+        beat = self.get_object()
+
+        input_ser = ContributeBeatInputSerializer(data=request.data)
+        input_ser.is_valid(raise_exception=True)
+        data = input_ser.validated_data
+
+        character_sheet_id: int = data["character_sheet_id"]
```
TehomCD: again, same problem

src/world/stories/views.py
```diff
@@ -339,3 +481,1029 @@ def feedback_given(self, request: Request) -> Response:
...
+        # Verify the requesting user owns this character_sheet (or is staff).
+        try:
+            character_sheet = CharacterSheet.objects.select_related("character").get(
+                pk=character_sheet_id
+            )
+        except CharacterSheet.DoesNotExist:
+            return Response(
+                {"detail": "Character sheet not found."},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
```
TehomCD: huge code smell to not use serializer for this

src/world/stories/views.py
```diff
@@ -339,3 +481,1029 @@ def feedback_given(self, request: Request) -> Response:
...
+
+        if not request.user.is_staff:
+            if character_sheet.character.db_account_id != request.user.pk:
+                return Response(
+                    {"detail": "You may only contribute for your own character."},
+                    status=status.HTTP_403_FORBIDDEN,
+                )
```
TehomCD: should be a permission class

TehomCD (changes requested):
The views have a repeat of problems we've had before: not properly using serializers or permission classes.

We should always have this pattern:
permission classes check who can call the view/action
serializers validate the data, raising validation errors if there's any invalid states
service functions are called with validated data, either directly from the serializer or the view, and can no longer raise errors from those invalid states

It's a huge code smell for service functions to get raw data and raise exceptions from it, and it's a huge code smell for views to be checking these things directly. The machinery in django rest framework exists to separate these concerns. Overloading views means we often are duplicating logic with unclear responsibilities

ApostateCD:
- All 15 action endpoints refactored to the canonical three-layer split:                                                                                                          
    - Permission classes (permissions.py): IsGMProfile, IsAccountOfCharacterSheet, Lead-GM-on-episode/beat/claim/session-request variants — every inline try:                       
  request.user.gm_profile block and inline role check is gone                                                                                                                           - Input serializers (serializers.py): PrimaryKeyRelatedField(queryset=...) for every FK; validate() / validate_<field>() for every business rule (agm_eligible check, status    
  transition check, character_sheet ownership, etc.); view-resolved objects passed via context                                                                                          - Service functions (services/): receive validated data only. StoryError subclasses removed from user-input-validatable paths. assert → if not ...: raise ValueError(msg) for     programmer-error guards (ruff S101)                                                                                                                                               
  - One surgical exception: EpisodeViewSet.resolve keeps a narrow try/except StoryError for NoEligibleTransitionError and AmbiguousTransitionError — these are genuinely
  race-condition errors (eligibility can change between serializer validation and service execution), explicitly called out in CLAUDE.md as the only permitted try/except in a view 
  body
  - src/world/stories/CLAUDE.md gains a canonical "Action Endpoint Pattern" section with code examples and an anti-pattern table — this is the documentation you asked for to       
  prevent the pattern recurring
  - 510/510 tests pass on both --keepdb and fresh DB

  Now when someone adds an action endpoint in this app, the first thing they'll see in the app's CLAUDE.md is the three-layer split spelled out with examples and anti-patterns. The
   global CLAUDE.md already had the rule — this makes it impossible to miss for stories work.

TehomCD (approved):

PR #410, opened 2026-04-30
Inventory service functions design

ApostateCD:
## Summary                                                                                                                                                                        
                                                                                                                                                                                    
  - Adds 7 inventory service functions in `flows/service_functions/inventory.py`
    (pick_up / drop / give / equip / unequip / put_in / take_out) composed on
    top of the items data layer that landed in #409
  - Adds `ItemState` flow object state with `_run_package_hook`-routed permission
    methods so behaviors/triggers can deny actions
  - Wires the existing `GetAction` / `DropAction` / `GiveAction` to use the new
    service functions (so they now track ownership and write `OwnershipEvent`
    rows on transfer); adds `EquipAction` / `UnequipAction` / `PutInAction` /
    `TakeOutAction`
  - Adds telnet `wear` / `remove` / `put` / `withdraw` commands; extends `CmdGet`
    so `take <item> from <container>` dispatches to `TakeOutAction`
  - Adds `execute_action` WebSocket inputfunc — the canonical mutation channel
    for the React frontend
  - Removes `EquippedItemViewSet` POST/DELETE + `EquippedItemWriteSerializer` +
    `EquippedItemWritePermission` (they bypassed the action layer's auto-swap
    policy); the ViewSet remains as read-only list/retrieve

  Establishes the rule going forward: **all mutations flow through the action
  layer (telnet via `Cmd*`, web via WS `execute_action`), REST is read-only.**

  Design doc: `docs/plans/2026-04-29-inventory-service-functions-design.md`
  Implementation plan: `docs/plans/2026-04-29-inventory-service-functions.md`

  ## Test plan

  - [x] `arx test world` (full regression) — 4951 tests pass
  - [x] `arx test world.items actions commands flows server.conf web` targeted
  - [x] All 25 commits land cleanly atop `e76f011c` (Items and Resonance #409)
  - [ ] Manual smoke test: telnet `wear cloak` → succeeds, item appears in
        EquippedItem table, room broadcasts pose
  - [ ] Manual smoke test: WS `execute_action` with `action=equip` → same result
  - [ ] Manual smoke test: `take ring from chest` works for chest in room and
        chest in inventory


src/actions/definitions/item_helpers.py
```diff
@@ -0,0 +1,26 @@
+"""Shared helpers for item-related actions.
+
+These helpers bridge the actions layer (which receives ``ObjectDB`` targets)
+to the ``world.items`` domain models that inventory service functions operate
+on.
+"""
```
TehomCD, line 6: I don't really like having a module name start with an underscore, as if it's a private module. I've never seen this convention before and I don't think we should adopt it. Private modules aren't a thing

src/commands/evennia_overrides/items.py
```diff
@@ -0,0 +1,110 @@
...
+    def resolve_action_args(self) -> dict[str, Any]:
+        args = (self.args or "").strip()
+        if not args:
+            msg = "Wear what?"
+            raise CommandError(msg)
+        target = self.caller.search(args, location=self.caller)
+        if not target:
+            msg = f"Could not find '{args}'."
+            raise CommandError(msg)
+        return {"target": target}
```
TehomCD: I feel like we could DRY out these functions a little bit, like have a class variable for the message for no args and have that automatically be checked/raised, for if we can't find the different objects, etc

src/flows/service_functions/inventory.py
```diff
@@ -0,0 +1,192 @@
...
+    If the item is currently unowned (``owner`` is null), ``character``'s
+    account becomes the owner. Pre-existing ownership is preserved.
+    """
+    if not item.can_take(taker=character):
+        raise PermissionDenied
+    item.instance.game_object.location = character.obj
+    item.instance.game_object.save()
+    if item.instance.owner is None:
+        item.instance.owner = character.obj.account
+        item.instance.save(update_fields=["owner"])
```
TehomCD, line 49: should we use move_to so that it fires Evennia hooks?

ApostateCD:
## Update: reachability + possession enforcement                                                                                                                                                                                                                                                                                                                        A final review pass found that `ItemState.can_*` methods were defaulting to                                                                                                       
  True with no actual validation, leaving the service functions vulnerable to                                                                                                         the WebSocket path bypassing them with arbitrary `*_id` values. Telnet was
  incidentally bounded by `caller.search()` scoping, but the WS path wasn't.                                                                                                        
                                                                                                                                                                                    
  Closes 6 gaps across the inventory ops:

  - `pick_up` accepted items from any room → now checks reachability
  - `drop` didn't verify the item was actually carried → now checks possession
  - `give` let the giver transfer items they didn't have → now checks possession
  - `equip` let the wearer equip items they didn't carry → now checks possession
  - `put_in` didn't validate the container was reachable → now checks reach
  - `take_out` had zero auth — opened any container anywhere → now full validation

  Single source of truth: `ItemState.is_in_possession(character)` and
  `is_reachable_by(character)` walk the `contained_in` chain (rejecting closed
  containers). The four `can_*` methods route through them. Service functions
  just call `can_*` and trust the answer. The WS dispatcher needed no change —
  service-layer validation closes the gap for both transports.

  Also added two typed exceptions (`NotReachable`, `NotInContainer`) so service
  functions raise the specific exception per failure mode instead of a generic
  `PermissionDenied`. `pick_up` and `drop` now also clear `contained_in` when
  moving items out of containers, so we never leave inconsistent FK + location
  state.

  8 new tests cover the previously-unguarded attack surface; full regression
  (4962 tests in `arx test world`) still green.

TehomCD (approved):

PR #416, opened 2026-05-03
Visible worn equipment

ApostateCD:
## Summary                                                                                                                                                                                                                                                                                                                                                              Looking at a character now shows their visible worn equipment with deeper                                                                                                         
  layers concealed by covering items. Drilling into a piece reveals its
  description on both transports — telnet uses `look bob's hat` /
  `look hat on bob` / `look coin in pouch`; the React frontend uses a
  focus-stack side panel with dynamic tab name (Room → character name →
  item name).

  Cross-cutting: this PR also lands a small staff-bypass infrastructure
  (`PlayerOnlyPermission`, `PlayerOrStaffPermission`, `is_staff_observer`)
  that the visibility service uses, and refactors three existing
  permission classes to use the new bases. The principle: staff bypass
  is explicitly opt-in per resource (picking the base class IS the
  opt-in), never automatic — flagged via the audit that the
  `CanViewInteraction` class has an intentionally non-uniform policy
  that excludes staff from very-private interactions.

  Design doc: `docs/plans/2026-05-02-visible-worn-equipment-design.md`
  Plan: `docs/plans/2026-05-02-visible-worn-equipment.md`

  ## Test plan

  - [x] Targeted backend regression — `arx test world.items actions
        commands flows core_management web --keepdb` passes
  - [x] Frontend `pnpm typecheck && pnpm lint` clean
  - [ ] Full no-keepdb regression on CI (the 2 stale-keepdb failures
        noted in `flows.tests.test_scene_data_manager` are pre-existing
        and unrelated — pk=999 cruft from prior runs; that test file is
        unchanged in this PR)
  - [ ] Manual: telnet `look <person>'s <item>` shows item description
  - [ ] Manual: frontend click character in room panel → drill into
        character focus, click worn item → drill into item focus,
        back-button returns
  - [ ] Manual: staff user can see concealed items (admin convenience)


src/world/items/views.py
```diff
@@ -392,3 +384,165 @@ def perform_destroy(self, instance: OutfitSlot) -> None:
...
+        qs = super().get_queryset()
+        if self.request.user.is_staff:
+            return qs
+        return qs.filter(id__in=self._visible_item_ids())
+
+    def _visible_item_ids(self) -> set[int]:
+        """Item IDs currently visibly worn on characters the user can observe."""
+        user = cast(AccountDB, self.request.user)
+        own_entries = RosterEntry.objects.for_account(user)
+        own_char_ids = list(own_entries.values_list("character_sheet_id", flat=True))
```
TehomCD: All of this is the wrong approach with SharedMemoryModels. Any related list should be cached, and all these objects should be derivable from cached_properties on the SharedMemoryModels themselves. In this format, we're making something like 6 + N queries every time someone is going to do a look command. That's ridiculous. We should be doing zero if these are all in-memory, which they should be the first time they're loaded.

TehomCD: Alternately, have some efficient subquery that the queryset can use. Having a ton of queries like this is crazy

src/world/items/views.py
```diff
@@ -392,3 +384,165 @@ def perform_destroy(self, instance: OutfitSlot) -> None:
...
+                continue
+            # Pick the right observer per character so the service applies
+            # the matching bypass: self-look (own character) → bypass; same
+            # room → no bypass (concealed items hidden).
+            observer: object
+            if character_id in own_char_ids:
+                observer = character
+            else:
+                observer = user
+            for entry in visible_worn_items_for(character, observer=observer):
```
TehomCD: `visible_worn_items_for` does a query with a prefetch. You're doing this in a loop. This is hideously inefficient and also totally unnecessary when we should have all this data in memory to begin with.

TehomCD (changes requested):
I think django rest framework viewsets are the right call for many things, but I'm starting to reconsider it when we'll be querying lists of objects that will be in memory due to SharedMemoryModels. In those cases, we shouldn't be querying the database at all: we should just be looking at the objects we need, retrieving them via in-memory processing, and serializing them. The Item views are especially egregious because they do a lot of queries in loops for data we already have

TehomCD (dismissed):

src/world/items/views.py
```diff
@@ -392,3 +386,185 @@ def perform_destroy(self, instance: OutfitSlot) -> None:
...
+        if item_pk is None:
+            raise NotFound
+
+        user = cast(AccountDB, request.user)
+
+        for character, observer in self._observable_characters_with_observer(user):
+            for visible in visible_worn_items_for(character, observer=observer):
+                if visible.item_instance.pk == item_pk:
+                    serializer = ItemInstanceReadSerializer(visible.item_instance)
+                    return Response(serializer.data)
```
TehomCD: This implementation is kind of nuts. We're querying all the roster entries in the game, checking if they match what we have, looking at any that have characters in the same room as our characters, looking at all their equipment, and if so... then we match the pk of the equipment and see if they match and return the data? This is crazy. Item being visible or not to our character might make more sense as a permission check, but either way we should be fetching the item -first-. The idea of looking up all the characters for our user and seeing which characters are in the same room as our characters and checking their items is bonkers. It's essentially a permission check for 'Can this account view that item'? The permission logic is if we're staff, yes. If we're not, is it visible to one of our characters? I think it's cleaner to pass in the character who is attempting to view it if it's essentially an inspection action: while we can derive the observer from all the characters the user has (which should in fact be cached, and not something we should be querying through RosterEntries), why not pass that in from the frontend and treating it as an action of an observer? But either way, the roster entry queries are unnecessary and strange. At that point, we just need to know what we can view and it's a check of it being our own character or a character we can see the gear of. either way there shouldn't need to be queries and "character with the item" should be directly derived from the item instance's location.

ApostateCD: ## Update: item-first refactor per senior review

  Senior reviewer pointed out the visible-worn endpoints had the wrong
  *shape* — searching all observable characters when the right pattern is
  fetch the item and treat visibility as a permission. Refactored both:

  - Both endpoints now require `?observer=<character_pk>` from the
    frontend (the active puppet doing the looking).
  - Detail endpoint fetches the item directly, derives the wearing
    character from `item.game_object.location`, and answers "can this
    user view this item?" as a permission check.
  - All `RosterEntry` queries removed from these views. The user-owns-
    observer check is `observer.db_account_id == user.id` — single FK
    attribute access, identity-map cached.
  - `_is_concealed_for_observer` walks `wearing_character.equipped_items`
    (cached handler) — no queries.

  Query count after refactor:
  | Endpoint | Before | After |
  |---|---|---|
  | `GET /visible-worn/?character=N&observer=M` | 3 | **1** |
  | `GET /visible-item-detail/<id>/?observer=M` | 3 | **2** |

  (All counts after handler warm-up. The remaining queries are unavoidable
  DRF auth + the initial `ItemInstance.objects.get(pk=...)`.)

  Frontend updated to pass `observerId` from the active puppet via the
  existing `MyRosterEntry.character_id` field. Focus views disable when
  no observer is available.

  The broader item-views audit (`ItemInstanceViewSet`, `EquippedItemViewSet`,
  `ItemFacetViewSet`, `OutfitViewSet`, `OutfitSlotViewSet` — all
  `ModelViewSet` subclasses doing similar DB queries when data is in
  memory) is tracked as a follow-up since refactoring those touches
  mutation semantics, not just reads.

TehomCD (approved):

PR #423, opened 2026-05-05
fix(character_creation): tradition list leaked beginning_id-filtered …

ApostateCD:
…data across requests

`TraditionViewSet.get_queryset` attached `prefetched_beginning_traditions` (filtered by `beginning_id` from the request) onto each Tradition via `Prefetch(to_attr=...)`. Tradition is a SharedMemoryModel — its instances persist in the per-process identity map across requests. Django's `prefetch_related` sees the to_attr is already set on every Tradition instance from the prior request and SKIPS the new prefetch, returning the previous request's filtered data.

Same shape as the codex tree leak (PR #418): two sequential requests with different `beginning_id`s on the same gunicorn worker would see each other's `required_distinction_id`. The list response varies by `beginning_id` because BeginningTradition rows differ per beginning, so the leak was directly user-visible.

The audit found this is the only other instance of the pattern in the repo. The remaining ~50 `Prefetch(to_attr=...)` sites are safe because their inner querysets are static — same answer for every caller — so the cache hit returns correct data.

Fix shape mirrors codex: pull the BeginningTradition rows in a separate flat query (one round trip), build a `{tradition_id: BT}` dict, and pass it via serializer context. Nothing per-request lands on the shared Tradition instance. Use a `Case`/`When` annotation to preserve BeginningTradition.sort_order on the Tradition queryset.

Adds `TraditionListLeakTests.test_required_distinction_is_per_beginning_not_per_process` as a regression guard: hits the endpoint with two different `beginning_id`s against the same Tradition and asserts each request returns its own data.

### DIFF
--- a/src/core/natural_keys.py
+++ b/src/core/natural_keys.py
@@ -49,6 +49,7 @@ from typing import TYPE_CHECKING, Any
 
 from django.db import models
 from django.db.models.fields.related import ForeignKey
+from evennia.utils.idmapper.manager import SharedMemoryManager
 
 if TYPE_CHECKING:
     from collections.abc import Sequence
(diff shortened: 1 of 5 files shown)


TehomCD (changes requested):
There's a few things wrong here. First, we should never cache something on the view. That's the wrong place for it. If we want to cache querysets, that should either be:

- a) in a SharedMemoryModel instance via a cached property: this would be for a related list of objects that point to it with a FK, usually with a handler for managing it

- b) in the SharedMemoryModel class instance: already does this with the dict, we could have other class methods for retrieving based on that

- c) in the SharedMemoryModel class `objects` manager. This is also a singleton that would support caching.

The example in here is especially bad because we're caching based on a request param that we're checking once for the first time we cache it and then we never update the cache. So if it was called again with a different value, it'd get the wrong data.

I would often change how we fetch things with a SharedMemoryModel to go to the root object we're filtering by, and fetch the related data which is often already cached, or should be cached. We're filtering by a `Beginning`. From that, we have its beginning traditions, and each of those has a tradition. So the viewset could be off the Beginnings viewset, which would prefetch all this data and return it for the given beginning we're interested in. All of it would cached, retrieved once, while we're doing a simple get by PK for beginning, which is also cached. If the entire viewset was called twice, it would perform zero queries, because a `Model.objects.get(pk=pk)` hits the cache, and then all the prefetches are in cached properties on the model.

TehomCD (approved):

PR #424, opened 2026-05-06
refactor(events): move active_persona_ids cache off the viewset onto …

ApostateCD:
…Account

TehomCD's review on PR #423 pointed out that caching state on the viewset is the wrong location even when the viewset is per-request fresh: the cache is discarded when the request ends, providing no value beyond deduplicating within a single request.

The architecturally correct cache anchor for "this account's active persona IDs" is the Account itself — Evennia's identity map keeps the same Account instance in memory across all requests in a process, so a ``cached_property`` on the typeclass survives the request boundary and is computed once per account per process.

Changes:

1. Add `Account.cached_active_persona_ids` (typeclasses/accounts.py). Direct join from AccountDB to Persona, no PlayerData hop required. Stale only on persona add/remove; resets on process restart.

2. `EventViewSet._active_persona_ids` is now a thin delegator to `request.user.cached_active_persona_ids`. The viewset stores nothing. Drops `RosterEntry`, `PersonaType`, and `cached_property` imports from the events views (no longer needed here).

3. `test_event_retrieve_query_count_with_gm_check` now locks the steady-state count at 3 queries (was 4). The persona-ids lookup that previously fired on each request is now a cache hit after the warmup. Updated the comment to explain the cache anchor.

### DIFF
--- a/src/typeclasses/accounts.py
+++ b/src/typeclasses/accounts.py
@@ -116,19 +116,25 @@ class Account(DefaultAccount):
         return CharacterList(self)
 
     @cached_property
-    def cached_active_persona_ids(self) -> list[int]:
-        """PRIMARY persona IDs for this account's currently-played characters.
+    def cached_primary_persona_ids(self) -> list[int]:
+        """IDs of PRIMARY personas this account is currently playing.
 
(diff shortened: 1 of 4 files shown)


TehomCD (approved):
lgtm

PR #442, opened 2026-05-13
Feature/combat viewset audit

ApostateCD:
Commit 1 (bcd5c6a1) — combat viewset audit
  - join requires explicit character_sheet_id; rejects auto-select pattern
  - Per-request roster cache via _viewer_character_ids
  - _current_round_action helper consolidates four duplicate lookups
  - New test_view_query_counts.py locks list/retrieve/my_action budgets

  Commit 2 (98360687) — adversarial review fixes + bundled follow-ups
  - Combat review fixes: permission classes route through the request cache (my_action budget drops 5 → 4), min_value=1 on the new serializer, documented
  .first() safety
  - Magic correctness fix: 4 ViewSets (CharacterAura/Resonance/Gift/Anima) were filtering by character__db_account=user (currently-puppeting), excluding
  alts. Now use RosterEntry.objects.for_account like everywhere else.
  - Flow test flake fix: 3 TestSceneDataManager tests used pk=999 as "doesn't exist," but the fallback ObjectDB.objects.get(pk=999) can hit a real row in
  cross-app fresh-DB runs. Mocked the lookup.
  - Items test pollution guard: CharacterCarriedItemsHandlerTests and CharacterSheetOutfitsHandlerTests failed in cross-app runs; added flush_instance_cache
   + invalidate in setUp matching the established pattern.

  Test results

  - 2155/2155 pass on the combined world.items + world.combat + world.magic + flows suite, both --keepdb AND fresh-DB
  - Pre-existing scene_data_manager flakes are now fixed in cross-app runs

  Skipped from the cleanup-bundle option

  - N2 (Outfit.character_sheet related_name rename) — migration risk, not worth bundling
  - noqa malformed-directive sweep — too broad (10+ files, mechanical), better as its own focused PR
  - Full test-pollution audit — variable scope, ad-hoc as discovered

  Open follow-ups (still in memory, not done)

  - drf-spectacular introspection fix for the new viewsets.ViewSet classes (blocks frontend type regen)
  - Staff scheduler tooling (needs brainstorm)
  - Item ViewSet audit follow-ups from earlier reviews

### DIFF
--- a/src/world/combat/permissions.py
+++ b/src/world/combat/permissions.py
@@ -11,6 +11,24 @@ from world.combat.models import CombatEncounter
 from world.roster.models import RosterEntry
 
 
+def _viewer_character_ids(request: Request, view: APIView) -> set[int]:
+    """Return the request user's played character_sheet ids.
+
+    Prefers ``view._viewer_character_ids(request)`` when the view exposes
+    it (``CombatEncounterViewSet`` does — it caches the set on the
+    request object so permission checks and view body share the same
+    roster query). Falls back to a direct query for any caller wired up
+    with these permissions but lacking the helper.
+    """
+    helper = getattr(view, "_viewer_character_ids", None)  # noqa: GETATTR_LITERAL
+    if helper is not None:
+        return helper(request)
+    if not request.user.is_authenticated:
+        return set()
+    user = cast(AccountDB, request.user)
+    return set(RosterEntry.objects.for_account(user).character_ids())
+
+
 class IsEncounterGMOrStaff(BasePermission):
     """Allow access to GMs of the encounter's scene or staff.
 
@@ -35,7 +53,9 @@ class IsEncounterParticipant(BasePermission):
     """Allow authenticated users who have an active CombatParticipant.
 
     Uses the encounter's participants_cached when available (prefetched
-    by _base_queryset) to avoid a separate query.
+    by _base_queryset) to avoid a separate query. Routes the roster
+    lookup through the view's per-request cache when available so the
+    permission check and view body share a single roster query.
     """
 
     def has_object_permission(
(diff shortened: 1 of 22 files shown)


src/world/combat/permissions.py
```diff
@@ -1,14 +1,22 @@
...
+def _viewer_character_ids(request: Request, view: APIView) -> set[int]:
+    """Return the request user's played character_sheet ids.
+
+    Always routes through ``view._viewer_character_ids(request)`` —
+    these permissions are wired to ``CombatEncounterViewSet`` (in
+    production) and to a stub view in ``test_permissions`` (in tests).
+    The view caches the set on ``request._combat_viewer_character_ids``
+    so the permission check and view body share a single roster query.
+    """
+    return view._viewer_character_ids(request)  # noqa: SLF001 — known cooperator
```
TehomCD: I really think this is the wrong place for the data to be cached. Ultimately, we're getting it from the account which is passed in from the view. We should have the character sheets associated with an account set as a cached property on the Account typeclass, which is what we should get on a request. If we don't have a proper Account object (an unauthorized/anonymous user account object, for example), then we can catch the AttributeError and return an empty list

src/world/combat/views.py
```diff
@@ -465,6 +466,25 @@ def flee(self, request: Request, pk: int | None = None) -> Response:
...
+            return request._combat_viewer_character_ids  # noqa: SLF001
+        if not request.user.is_authenticated:
+            ids: set[int] = set()
+        else:
+            user = cast(AccountDB, request.user)
+            ids = set(RosterEntry.objects.for_account(user).character_ids())
+        # Per-request cache attribute. Namespaced with the combat prefix so
+        # it doesn't collide with any other code stashing state on request.
+        request._combat_viewer_character_ids = ids  # noqa: SLF001
+        return ids
```
TehomCD: Wrong place for this, it should be cached on the account object and if our request user isn't an instance of our Account typeclass/doesn't have the property, return an empty list

TehomCD (approved):

PR #446, opened 2026-05-16
Feature/stories authoring framework design

ApostateCD:
Ships: the runtime authoring backbone — StoryMaturity/BeatKind/ProgressStatus/UNASSIGNED enums, model fields + 2 additive migrations, maturity-promotion
  validation, frontier resolution (WAITING_FOR_GM/RESTING, routing-block distinction, atomic status reconcile), scope-assignment guard, Beat risk
  trust-gate, append-only StoryNote API (canonical 3-layer), dashboard staleness surfacing. Every richer beat resolves via the existing GM-mark placeholder;
   all deferred-engine seams in place, no schema churn later.

  Explicitly deferred (recorded in the design/plan docs, out-of-scope #1–#8): the product authoring API/UI for the new fields + the StoryAuthorPage reshape
  (I-1); a pitch text field + maturity-gated player visibility (I-2); the Mission/Challenge, Situation/Encounter+Sessions, reward-computation, GM-leveling,
  and covenant-entity engines. So this is the engine, runnable end-to-end at the service layer — staff/GM author-and-run through a UI is the next sequenced
  brainstorm.


TehomCD (approved):
The queries in stories/views are kind of concerning (usually a pretty bad sign when you see a view make that many queries) but the number of stories will be a pretty small table, so I think it'll be alright

PR #447, opened 2026-05-17
Feature/stories authoring api UI

ApostateCD:
Stories Authoring API + minimal-functional UI

  Makes the merged stories-authoring backbone usable end-to-end: staff/GMs can author and run a story through the browser, and the GMQueue/StaffWorkload
  query debt is paid down.

  Delivers
  - Backend: Story.summary (+migration 0031); maturity/summary/resting_conclusion/is_ending exposed on Story/Chapter/Episode serializers (incl. create);
  role-gated GM↔player text split (description/consequences GM-only, summary = player "The Story So Far", PITCH-blanked, fail-closed default-deny); POST
  /api/episodes/{id}/promote/ (PLOT-gate→400) and POST /api/stories/{id}/assign-to-scope/ (UNASSIGNED→character/group/global, creates progress, rejects
  re-assign); GMQueue/StaffWorkload bounded to constant query counts (assertNumQueries-locked, status-agnostic GM membership preserved).
  - Frontend: reshaped StoryAuthorPage — Beat kind/advances/risk (risk staff-gated), "Internal GM Description"/"The Story So Far" labels, episode
  resting_conclusion/is_ending, maturity-promote control (inline gate error), scope-assign dialog, GM Notes tab, inline progress banner, run-control
  (Resolve/Mark) + nimble +Beat/+Branch all on the author page.

  Resolves backbone follow-ups: I-1 (no authoring API), I-2 (GM/player text split — description=GM pitch, summary=player recap, no separate pitch field by
  design), #2/#3 (unbounded scans + PLR0913 dataclass).

PR #449, opened 2026-05-18
Feature/stories authoring cleanup

ApostateCD:
Body:
  ## Summary
  Clears every deferred low/no-design follow-up the merged stories authoring
  API/UI PR (#447) left in the framework-design ledger. No schema/migration
  changes; no scope creep.

  - (a) shared `episode_meets_plot_gate()` predicate (DRY, behavior-neutral)
  - (b) `IsLeadGMOnStoryOrStaff` isinstance routing; dead branch removed
  - (c) `_collect_gm_queue` docstring softened to accurate wording
  - (d) `StoryStatus.ACTIVE` over raw `"active"` literal (query-identical)
  - (e) `progress_status` exposed on `/api/stories/my-active/`; banner
    distinguishes waiting-for-GM vs resting (no new endpoint)
  - M-1 story owners privileged in `_gm_text_gate` (read back own GM text)
  - M-2 deliberate-ungated-echo SECURITY comment on Create serializers
  - M-3 `ScopeAssignDialog` Array.isArray-guarded error fields
  - I-A twin `markBeat` attaches failed `.response` for `MarkBeatDialog`

  Ledger updated with per-item RESOLVED notes + one noted query follow-up.

  ## Test Plan
  - [x] Full `world.stories` on a fresh DB: `Ran 881 OK` (879 baseline + 2 new)
  - [x] backend `ruff check` / `ruff format --check` / `ty check views.py` clean
  - [x] FE `pnpm build` (tsc -b + vite) clean
  - [x] FE vitest 80/80 across all 8 touched stories test files
  - [x] Adversarial review: SHIP (no Critical/Important)


src/world/stories/serializers.py
```diff
@@ -166,7 +166,16 @@ def _gm_text_gate(
...
-    if role not in (VIEWER_ROLE_STAFF, VIEWER_ROLE_LEAD_GM):
+    # Story owners are privileged for their own story's GM text (ledger M-1):
+    # a non-staff, non-lead-GM owner could not previously read back their own
+    # description/consequences after a PATCH (friction, not a leak). One
+    # .exists() only on the otherwise-stripped path.
+    is_owner = (
+        user is not None
+        and getattr(user, "is_authenticated", False)  # noqa: GETATTR_LITERAL — AnonymousUser safe
+        and story.owners.filter(pk=user.pk).exists()
+    )
```
TehomCD: stories should cache their owners in a property and we should fetch the set of pks from that and use it to compare rather than do a query. or have a method in the story that does this. Either way, we shouldn't do a query every time here for something that should be cached

src/world/stories/serializers.py
```diff
@@ -166,7 +166,16 @@ def _gm_text_gate(
...
+    if role not in (VIEWER_ROLE_STAFF, VIEWER_ROLE_LEAD_GM) and not is_owner:
```
TehomCD: thinking about it, it bugs me that we're having a function that explicitly is checking access for the user and then we just sort of say "jk, we're having a separate check here". the only thing the roles it returns are used for is this permission check. Why not have a "can_see_this_stuff" function instead? Or a role that says as much if they're the owner? This seems redundant

src/world/stories/services/maturity.py
```diff
@@ -12,6 +12,21 @@
...
+    Single source of truth for the PLOT promotion rule, shared by
+    ``promote_episode_maturity`` (service, raises) and
+    ``PromoteEpisodeInputSerializer.validate`` (Layer-2, 400). The rule:
+    non-empty ``resting_conclusion`` AND (an outbound transition OR
+    ``is_ending``). Independent of direction — callers decide when the gate
+    applies (upward move *to* PLOT only).
+    """
+    if not episode.resting_conclusion.strip():
+        return False
+    return episode.outbound_transitions.exists() or episode.is_ending
```
TehomCD: better to have the non-query first so it short-circuits and avoids the query if it's True

TehomCD (dismissed):
Optional changes to serializer if you like - I felt some of the checks should have been elsewhere

src/world/stories/models.py
```diff
@@ -186,6 +186,18 @@ class Meta:
...
+    def owner_account_ids(self) -> frozenset[int]:
+        """Cached set of owning AccountDB pks.
+
+        One query on first access, then reused for the life of this
+        identity-mapped Story instance (so the GM-text gate does not
+        issue an owners query per serialization). Owners change rarely;
+        a stale set after an in-process m2m mutation is acceptable for
+        this read-path and consistent with the SharedMemoryModel cache.
+        """
+        return frozenset(cast(Any, self).owners.values_list("pk", flat=True))
```
TehomCD, line 202: This is good but like other handlers we add, we should invalidate if it changes - it'd be good to add something to the views where we add/remove owners to invalidate this cache

TehomCD (approved):

PR #452, opened 2026-05-19
Turnkey infra-as-code stand-up button + observability baseline (latent)

TehomCD:
## Summary

Adds the **turnkey production stand-up "button"** (OpenTofu + Ansible, one gated click) plus the **Plan-1 observability exporter** (off-by-default). Large but **intentionally latent** — nothing runs until the LLC/Linode/Cloudflare/Resend accounts exist and the gated workflow is manually pressed.

- `infra/` — bootstrap state bucket; OpenTofu modules (compute, linode_firewall, cloudflare_dns incl. Resend/SPF/DKIM/DMARC, object_storage, r2_offsite) + prod & isolated ephemeral-stage roots (full `prevent_destroy`, apply-only); 16 Ansible roles wired in `site.yml` (hardening, secrets-via-env, Postgres-localhost, Caddy DNS-01, TLS-telnet, Django hardening, deploy, 3-2-1 backups, alerting).
- The button: `infra/scripts/standup.sh` + repo-root `.github/workflows/standup.yml` (`workflow_dispatch`, gated `prod` Environment, DRY thin wrapper). Separate gated `restore.sh` + `restore-rehearsal.sh`.
- `infra/scripts/acceptance.sh` static safety harness (**29/29 PASS**); CI `validate.yml` (tofu/ansible-lint/caddy/shellcheck/**gitleaks**) is the authoritative dynamic gate.
- `src/evennia_extensions/observability/` — Plan-1 exporter, **off by default**, isolated tests passing.

## Honest status — do not merge until

- [ ] **`validate.yml` green.** Expect it **RED on first run**: `tofu`/`ansible-lint`/`caddy` have never executed against this code (authored on Windows where they can't run) — provider-schema/`tofu fmt`/lint fixups are anticipated and flagged in-code. This is the validation gate doing its job; scoped to `infra/**` only, won't block other checks.
- [ ] **Pin GitHub Action SHAs** in `validate.yml` + `standup.yml` (currently mutable tags; the one explicit pre-trust item from the focused security review — supply-chain hardening for the secret-touching gated workflow).
- [ ] **Account-time checklist** (printed by `acceptance.sh`): 2nd-apply-no-changes, ansible idempotence, external port scan, restore-rehearsal from both 3-2-1 copies — run once accounts exist.

## Security

- No secret material in the repo (verified: acceptance, manual scan; design = secrets only in the gated GitHub Environment, reach the host as env → a 0600 EnvironmentFile, never repo/state/logs/argv).
[...]

PR #478, opened 2026-05-20
Make the deploy button actually run Evennia: uv install, migrate, superuser, fix systemd

TehomCD:
## Summary

Follow-up to #452. The infra in #452 stood up the box, hardened the OS,
created an empty Postgres DB, and installed a systemd unit pointing at
`evennia start` — but nothing installed Python deps, ran migrations, or
created a superuser, and `evennia` wasn't on PATH because no venv existed.
The unit would have failed on first start with `evennia: command not found`.

This PR closes that gap so the button actually delivers a running game.
Also folds in two pre-existing `IMPLEMENTATION-VERIFY` items found while
investigating: `base_packages` was missing `git` (used by `app_deploy`'s
`ansible.builtin.git`), and `app_pidfile` pointed at `current/server/server.pid`
rather than the gamedir-relative `current/src/server/server.pid`.

Includes a small README cleanup committed locally during the #452 work but
never pushed: SSH-key bootstrap mini-section + scrub of the stale
ansible-vault prose (superseded by the gated-Environment-only model).

## What's in this PR

- **`base` role:** install Astral's `uv` from upstream tarball, pinned to
  v0.11.15 with sha256 verification, idempotent (skips re-download when
  already at the pinned version). Add `git` to baseline packages.
- **`app_deploy`:** after the atomic symlink and before systemd start, run
  `uv sync --frozen --no-dev` → `evennia migrate --noinput` → `evennia
  collectstatic --noinput` → superuser-exists guard → `evennia
  createsuperuser --noinput` (first run only). Every task is idempotent.
- **systemd unit:** `ExecStart`/`Reload`/`Stop` now go through
  `uv run evennia <verb>` so the venv binary resolves. `WorkingDirectory`
  is the gamedir. `app_pidfile` fixed.
- **`secrets_vault`:** map `ARXII_DJANGO_SUPERUSER_PASSWORD` →
  `DJANGO_SUPERUSER_PASSWORD` in the rendered 0600 EnvironmentFile.
- **`standup.yml`:** wire the new superuser secret (password) and
  Variables (username, email; sensible defaults if unset) through to the
  ansible step's env only.
- **README:** SSH-key bootstrap section + scrub stale ansible-vault prose
  (from the unpushed local commit), plus a new "What the button actually
  does to game state (first run vs. re-run)" section so a less-technical
  reader understands the migrate / superuser / idempotency story.
- **`acceptance.sh`:** 10 new static invariants (uv pinned + sha-verified,
  uv sync present, migrate/collectstatic present, superuser create
  exists-guarded, secrets map carries the new password, no plaintext
  password committed anywhere, systemd uses `uv run`, pidfile path fixed).

## Test plan

- [x] `tofu fmt -check -recursive` clean
- [x] `shellcheck infra/scripts/*.sh` clean
- [x] `infra/scripts/acceptance.sh` — all 39 static checks PASS (incl. 10
      new first-run-setup invariants)
- [ ] CI `Infra Validate` (tofu/ansible-lint/caddy/gitleaks) green
- [ ] arxii's existing CI (backend, frontend, pre-commit, ty, qodana,
      CodeQL) green — no application code touched, expect no regression
- [ ] **Account-time** (cannot be tested without Linode account, but
      worth running once it exists): one full end-to-end stand-up to
      confirm `uv sync` → `migrate` → `createsuperuser` → `evennia start`
      actually delivers a running, web-reachable, login-able game

## Out of scope (deliberately)

- Frontend (React/Vite) build/deploy — arxii-specific concern; the
  backend + Django admin will work without it. Telnet players can
  connect; web players need the frontend deployed (its own follow-up).
- Game-content fixtures — per maintainer, loaded via Django admin after
  the superuser exists; application-layer concern, not the deploy's job.


ApostateCD (approved):
lgtm

PR #491, opened 2026-05-21
Feature/test speedups

ApostateCD:
Two-tier testing (SQLite inner loop + Postgres parity) + CI shard hygiene

  Summary

  Local test runs were taking 30+ minutes for the full suite. This PR delivers:

  1. CI fix: 8 apps weren't placed in any CI shard, so their tests weren't running on PRs. Added the missing placements + a pre-commit linter that prevents
  the next new app from missing.
  2. Local inner loop: Adds a SQLite-in-memory test tier (just test-fast <app>) for the 14 apps where it works cleanly. Per-app runs that took 5-30s on
  Postgres now take 1-15s on SQLite. Postgres parity tier (every CI run + just test-parity locally) is unchanged.
  3. Test-quality fixes: ~25 tests had latent isolation bugs that Postgres's deferred FK semantics masked; the SQLite tier surfaced them. Fixes them in
  place where it was a real test bug, decorates with @tag("postgres") where the test genuinely needs PG.

  Honest up-front: the initial vision (from the senior dev) was full-suite SQLite at ~30 seconds. That didn't pan out for this codebase (details below); we
  landed on per-app SQLite for the dev's daily working set instead. The full-suite parity path stays on PG.

  ---
  Why this work, and the path we took

  Local fresh-DB test runs were taking ~33 minutes serial / ~15 min parallel for the full suite. CI is fine — sharded across 4 jobs on tmpfs Postgres,
  finishes in reasonable time. The friction is purely local-dev.

  Initial hypothesis (from the senior dev): switch local tests to SQLite in-memory, expect ~30 seconds full-suite, keep PG-only in CI. I pushed back too
  hard on this initially based on a literal reading of CLAUDE.md's "PostgreSQL Only" rule; the senior dev was right that the rule was intended for
  production code, not the test tier. Pivoted to the two-tier model after that correction.

  Reality vs the initial vision. Full-suite SQLite on this codebase doesn't hit 30 seconds. Three structural reasons surfaced during the spike:

  1. 5 apps have raw PG-only schema operations (materialized views in societies/codex/areas, range-partitioned tables in scenes, covenant mat-view in
  societies/0003). SQLite can't replay those migrations. Wrapped the offending RunSQL ops in a PostgresOnlyRunSQL subclass that no-ops on non-PG backends.
  2. SQLite can't replay column-drop-with-constraint-reorder migrations that Postgres handles natively. Several apps have this pattern from past schema
  refactors. Used MIGRATION_MODULES = DisableMigrations() to build the test schema directly from current model state instead of replaying migrations —
  sidesteps the issue entirely without per-migration archaeology.
  3. SharedMemoryModel's identity map is process-global, not transaction-scoped. Test classes can pull idmap-cached rows from earlier classes, which
  Postgres masks via different transaction isolation. SQLite's per-test FK check at teardown catches the resulting dangling references. This surfaced ~1,400
   errors in the full-suite SQLite run that don't appear in per-app runs.

  After surveying 19 apps individually, the working set looks like this:

  ┌───────────────────┬────────────────────────────────────────────────────────────┬────────────────────────────────────────────────────────────────────┐
  │       Tier        │                            Apps                            │                                Why                                 │
  ├───────────────────┼────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────┤
  │ SQLite-clean (14  │ action_points, forms, achievements, game_clock, skills,    │ Shallow fixture graphs; engine-style apps. Per-app run 0.3-15s on  │
  │ apps)             │ relationships, flows, checks, events, missions, journals,  │ SQLite vs 5-30s on PG.                                             │
  │                   │ mechanics, progression, conditions                         │                                                                    │
  ├───────────────────┼────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────┤
  │ SQLite with       │ combat, items, vitals, actions                             │ 5-15% of tests broken on SQLite. Not decorated in this PR — usable │
  │ caveats (4 apps)  │                                                            │  with failures visible until someone wants to invest.              │
  ├───────────────────┼────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────┤
  │                   │ roster, character_sheets, magic, scenes, codex, areas,     │ Fixture chains pulling the full Character → CharacterSheet →       │
  │ PG-only (7 apps)  │ societies                                                  │ Persona → RosterEntry → Account graph hit PG-vs-SQLite FK          │
  │                   │                                                            │ semantics; OR app has raw PG schema. Use just test-parity.         │
  └───────────────────┴────────────────────────────────────────────────────────────┴────────────────────────────────────────────────────────────────────┘

  ---
  What's in this PR (8 commits, by phase)

  Phase 1 — CI shard hygiene

  Benefits the codebase regardless of who runs tests locally — every contribution to the unsharded apps was silently uncovered by CI before this.

  - 596ffa65 — ci: add world.missions to shard-1. The original trigger from the missions PR review.
  - 44509d49 — feat(tools): add lint_shard_coverage to catch unsharded apps. A pre-commit linter that walks src/ for apps.py files, parses
  .github/workflows/ci.yml, fails on missing or duplicated apps. Key finding from running it against the branch: 7 apps beyond missions were also unsharded
  (world.locations, world.traits, world.areas, world.instances, evennia_extensions, world.realms, world.species). Added them to appropriate shards via
  smallest-first heuristic.

  Phase 2 — SQLite test tier scaffolding

  - 70b50dd2 — perf(test): SQLite in-memory test tier scaffolding. Five components: sqlite_test_settings.py with DisableMigrations sentinel, Windows worker
  shim, --sqlite flag on arx test (+ --exclude-tag), PostgresOnlyRunSQL helper in core_management/migration_utils.py, 5 PG-only migrations wrapped (no PG
  behavior change). Default test settings remain test_settings (PG); CI is unaffected.
  - e3b718e7 — perf(test): vendor-guard refresh_*_view() helpers for SQLite tier. Three service functions (refresh_legend_views, refresh_codex_breadcrumbs,
  refresh_area_closure) executed raw REFRESH MATERIALIZED VIEW — fixed ~233 distinct test errors by early-returning on non-PG.
  - 05085e28 — docs(perf): per-app SQLite survey across 19 apps. The working/broken set, in docs/perf/sqlite-spike-2026-05-20.md.

  Phase 3 — Test cleanup

  - bdd539da — test(world): decorate PG-required tests + flush SharedMemoryModel caches. ~25 tests across 6 apps. Each classified as @tag("postgres")
  (genuine PG dependency: materialized views, DISTINCT ON, Evennia DbHolder copy issues) OR fixed in place (SharedMemoryModel idmap flushes via
  flush_instance_cache() in setUp). Per-app results: events 1→0, journals 1→0, missions 4→0, mechanics 5→0, progression 6→0, conditions 8→0.

  - The 4 missions failures turning out to be idmap pollution (rather than lines[0]-without-order_by as initially diagnosed) is a real test-quality finding
  — those tests were latently broken; PG just gave them the right answer most of the time. They're now properly isolated regardless of engine.

  Phase 4 — Ergonomics

  - 2a30cdef — perf(test): just test-fast/test-parity recipes + restore arx-test-scratch.sh. Two new justfile recipes: test-fast (SQLite + exclude-tag
  postgres, serial — --parallel REGRESSED missions by 25% due to per-worker DB-clone overhead) and test-parity (PG, parallel). Also restored
  arx-test-scratch.sh which the existing recipe referenced but was missing.

  Phase 7 — Docs

  - e6c60440 — docs(CLAUDE.md): two-tier test model + working app set. "PostgreSQL Only" rule now explicitly scoped to production code; tests use the
  two-tier model.

  ---
  Anticipated questions

  Q: Why not just fix the SQLite-unfriendly migrations directly instead of DisableMigrations?

  A: Per-migration archaeology with no production benefit. world.magic alone has ~13 migrations touching the room_aura_profile field rename.
  DisableMigrations is the conventional Django pattern for "build test schema fresh from models" — production unaffected, PG tier still runs the full chain,
   arx manage migrate against real PG works exactly as before.

  Q: Why not switch arx test default to SQLite?

  A: CI runs arx test --parallel <shard apps> without an engine flag — switching the default would silently make CI use SQLite, breaking the parity tier.
  Keep arx test as PG; just test-fast opts in explicitly.

  Q: What about the 1,400+ errors in the full-suite SQLite run?

  A: Variants of SharedMemoryModel identity-map pollution + cross-app fixture FK timing — Postgres masks both. Could chase with a test-runner shim that
  flushes flush_instance_cache(force=True) between classes, but substantial work for marginal value (full-suite SQLite isn't viable anyway since several
  apps need migrations to exist). Per-app SQLite is the realistic win.

  Q: Should the SQLite tier be in this repo at all if it doesn't work for the full suite?

  A: just test-fast world.missions going from 18.8s to 5.3s is a meaningful daily improvement for the dev who works on missions/items/conditions/etc.
  Maintenance cost is small (~6 new files, ~80 lines in CLAUDE.md). The CI shard hygiene in Phase 1 is independently valuable; the SQLite work rides along.
  If after using it for a few weeks the carve-out feels worse than just running PG, we can rip it out — but the infrastructure is small enough that ripping
  out is cheap.

  Q: Why is PostgresOnlyRunSQL in core_management/migration_utils.py?

  A: core_management/ is the existing home for project-wide Django infrastructure (check_migrations.py, etc.). Natural neighborhood. Happy to move it if
  you'd prefer.

  Q: How does this interact with CI?

  A: It doesn't change CI behavior. CI's invocation doesn't pass --sqlite — still uses test_settings (PG). Only CI changes: shard assignments now include
  the 8 previously-uncovered apps, and the new shard-coverage pre-commit hook runs on every PR via the existing pre-commit job.

  Q: The 4 "partial" apps (combat, items, vitals, actions) — why not finish tagging them?

  A: Scope. Each would need per-test classification; volume is higher than the 6 we did in one pass. Left them at current --sqlite behavior (5-15% failures
  visible) so the next time someone works on them, they can decide to invest the cleanup or stick with just test-parity.

  Q: What's the failure mode if someone runs just test-fast world.character_sheets?

  A: Loud and obvious — 190/202 failures. They'll immediately see SQLite isn't viable for that app and either tag/fix or switch to just test-parity.
  CLAUDE.md + the survey doc list which apps work.

  ---
  What's NOT in this PR

  - PG-tier migration squashing. Audit landed (docs/perf/squash-audit-2026-05-20.md); no actual squashes. Worth a separate follow-up branch if PG-tier
  slowness becomes painful.
  - Persistent PG template DB. Same rationale.
  - Local tmpfs Postgres. Could be wired via docker-compose.test.yml if local just test-parity needs to match CI's speed.
  - setUpTestData audit sweep. Broader test-isolation work beyond what Phase 3.2 covered.
  - Tagging the remaining 4 "partial" apps.
  - Cross-app idmap test-runner shim for full-suite SQLite.

  ---
  Test gates

  - Per-app SQLite for the 14 clean apps: each runs OK via just test-fast world.<app>.
  - PG tier (CI parity): branch CI run on push will reproduce the existing 4-shard matrix, with 8 newly-placed apps now exercised.
  - Pre-commit linter smoke-tested: GREEN on current state; RED if world.missions removed from shard-1 OR world.combat duplicated into shard-2.
  - ruff + ty clean on every touched file.


src/core_management/migration_utils.py
```diff
@@ -0,0 +1,85 @@
...
+        self,
+        app_label: str,
+        schema_editor: BaseDatabaseSchemaEditor,
+        from_state: ProjectState,
+        to_state: ProjectState,
+    ) -> None:
+        """Run the reverse SQL only on Postgres; no-op everywhere else."""
+        if schema_editor.connection.vendor != _POSTGRES_VENDOR:
+            return
+        super().database_backwards(app_label, schema_editor, from_state, to_state)
```
TehomCD: Not sure why we need this if we're disabling migrations when using sqlite

docs/perf/test-baseline-2026-05-20.md
```diff
@@ -0,0 +1,47 @@
...
+
+CI runs each shard with `arx test --parallel <shard apps>` against a tmpfs Postgres
+(`ci.yml:80-91`). Local does NOT have tmpfs; this baseline therefore represents the
+worst-case (disk-bound) local-dev path.
+
+Shard test counts as of `44509d49` (per `ci.yml:46-76`):
+- shard-1: ~1,461 (magic 887 + combat 328 + missions 246)
+- shard-2: ~1,465 (stories 759 + mechanics 258 + conditions 190 + items 198 + areas 39 + instances 21)
+- shard-3: ~1,604 (flows 259 + progression 255 + scenes 227 + character_sheets 202 + character_creation 200 + gm 120 + actions 115 + locations 187 + traits 39)
+- shard-4: ~1,447 (all remaining apps; see ci.yml:67-76)
```
TehomCD: these perf md files all seem ephemeral records of tests, could probably wipe them

.claude/scripts/arx-test-scratch.sh
```diff
@@ -0,0 +1,56 @@
...
+PROJECT_ROOT="$(cd "$SCRIPT_DIR/../.." && pwd)"
+SCRATCH_DIR="$PROJECT_ROOT/.claude/scratch"
+OUT_FILE="$SCRATCH_DIR/$NAME"
+
+mkdir -p "$SCRATCH_DIR"
+
+# Auto-confirm Evennia's destroy-test-DB prompt by piping "yes" upfront.
+# tee captures combined stdout+stderr; pipefail propagates the runner's exit.
+echo "yes" | uv run arx test "$@" 2>&1 | tee "$OUT_FILE"
+exit "${PIPESTATUS[0]}"
```
TehomCD: was this a one-off test script?

src/world/areas/migrations/0002_create_areaclosure_view.py
```diff
@@ -22,7 +24,7 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrations.RunSQL(
+        PostgresOnlyRunSQL(
```
TehomCD: we may not need to do this if we disable migrations when running in sqlite

src/world/areas/models.py
```diff
@@ -86,6 +88,13 @@ def __str__(self):
...
 def refresh_area_closure() -> None:
-    """Refresh the areas_areaclosure materialized view."""
+    """Refresh the areas_areaclosure materialized view.
+
+    No-ops on non-Postgres backends — the view is PG-only (created via
+    PostgresOnlyRunSQL in migrations) and doesn't exist on the SQLite
+    inner-loop test tier.
+    """
+    if connection.vendor != _POSTGRES_VENDOR:
+        return
```
TehomCD: I sort of hate custom code to support tests being injected into functions here. I'd really prefer if instead we had a custom test runner for sqlite that handled monkey-patching where necessary to make these things no-ops

src/world/codex/models.py
```diff
@@ -155,7 +157,14 @@ class Meta:
...
 def refresh_codex_breadcrumbs() -> None:
-    """Refresh the codex_subjectbreadcrumb materialized view."""
+    """Refresh the codex_subjectbreadcrumb materialized view.
+
+    No-ops on non-Postgres backends — the view is PG-only (created via
+    PostgresOnlyRunSQL in migrations) and doesn't exist on the SQLite
+    inner-loop test tier.
+    """
+    if connection.vendor != _POSTGRES_VENDOR:
+        return
```
TehomCD: same here

TehomCD (commented):
I like the tags for tests that are postgres only. What I don't like is tweaks to migrations and model files that are based on which test database we're using.

In general I think we should have a custom test runner that would patch the classes it needs to, and since sqlite shouldn't be running migrations at all, I wouldn't think we need to change those. I could be wrong there but my expectation would be that we skip migrations so we don't care what's in them

.claude/scripts/arx-test-scratch.sh
ApostateCD: yeah just moved that local

TehomCD (approved):

PR #494, opened 2026-05-24
Feature/missions phase b

ApostateCD:
Body:

  ## Phase B — Mission authoring tooling model extensions

  Implements Phase B of the mission authoring tooling plan
  (`docs/plans/2026-05-22-mission-authoring-tooling-implementation.md`):
  the data-model shape that the Phase D DRF API + Phase E React Studio
  will eventually consume. Engine behaviour is unchanged — every field
  this PR adds is either "STORED BUT UNCONSUMED" until Phase D, or
  participates only in the new B7 audience-tier gate that staff bypass.

  ### What's in this PR (13 commits, 9 migrations)

  | Batch | Commit  | Summary                                                                           |
  | ----- | ------- | --------------------------------------------------------------------------------- |
  | B1    | 671f405d | `MissionCategory` lookup + `MissionTemplate.categories` M2M                       |
  | B2    | 8d2297da | `MissionGiver.giver_kind` discriminator + typed-FK fields + `MissionGiverOffering` through-model |
  | B3    | 428b9995 | `MissionGiverCooldown` → `MissionGiverStanding` (full rename + standing dimension) |
  | —     | 1f45e2dd | B1–B3 second-review polish                                                        |
  | —     | 0f0ced4f | Hand-fixed 0014 to apply on PG (M2M `through=` swap needs `RemoveField`+`AddField`, not `AlterField` — saved as memory) |
  | B4    | f9c322e2 | `MissionOptionRouteCandidate` per-candidate consequence + `outcome_text`; `MissionOptionRouteReward` route XOR candidate (DB
  `CheckConstraint`) |
  | B5    | 1c9614ab | `MissionNode.editor_x` / `editor_y` for Phase E canvas layout                     |
  | B6    | ed470da3 | `flavor_text` + per-field `*_needs_rewrite` flags for the Phase D copy service    |
  | —     | 4bd07ac2 | Batch-2 second-review polish (XOR `CheckConstraint`, STORED-BUT-UNCONSUMED inventory) |
  | B7    | 4b46378f | `MissionTemplate.access_tier` audience gate (deviation — see below)               |
  | —     | cf97ad48 | B7 review follow-ups (stale `publish()` doc refs, deploy note, factory comment)   |

  ### B7 deviation (called out in the docs)

  The original plan called for a full draft/publish working-copy fork
  (separate draft + live `MissionTemplate` rows, `publish()` to promote,
  `open_for_edit()` to fork). The user redirected to a simpler shape:
  a single `MissionTemplate.access_tier` audience field (`OPEN` /
  `STAFF_ONLY`, default `STAFF_ONLY`). Per-author in-flight protection
  is already provided by `MissionNodeSnapshot` (every accepted mission
  pins its node graph), so the remaining need is just "let staff test
  an authored mission before players see it" — handled cleanly by
  audience gating without forking the graph.

  `offer_missions` excludes `STAFF_ONLY` templates from non-
  `is_staff_observer(character)` characters; staff see both tiers;
  existing predicate / cooldown / level-band / arc-scope filters apply
  unchanged for everyone. The enum is intentionally minimal —
  richer tiers (society, GM-level, distinction-gated, etc.) defer to a
  dedicated permission brainstorm.

  Plan doc entries for B7 and D4 are updated with explicit DEVIATION
  callouts so the Phase D executor doesn't accidentally implement
  publish-fork services that no longer exist. D4 is now scoped to a
  `PATCH` action that flips `access_tier` (with an `is_publishable`
  guard before the `OPEN` promotion).

  ### Test plan

  - ✅ `just test-fast world.missions` — 301 passed (1 skipped)
  - ✅ `just test-parity world.missions` — 301 passed (1 skipped)
  - ✅ `just regression` (full no-keepdb, mirrors CI) — 7588 tests passed
  - ✅ Ruff clean; ty clean; pre-commit clean

  ### Notes for the reviewer

  - **Storage shape vs runtime semantics:** several fields land in B
    (e.g., `MissionOptionRouteReward.candidate`, `MissionGiverOffering.requirements_override`)
    but stay unconsumed by the engine until Phase D. Comments + the
    `STORED BUT UNCONSUMED` inventory at the top of
    `world/missions/services/resolution.py` track which fields the
    engine doesn't read yet.
  - **`MissionGiver.is_publishable`** is a property used only by the
    authoring/admin layer today; runtime enforcement in `offer_missions`
    is deferred to Phase D (would need either a factory refactor to
    populate a real location, or per-test updates to make 30+ existing
    tests pass with the gate). Roadmap deviation note records this.
  - **DEPLOY NOTE:** migration `0021` defaults every pre-existing
    `MissionTemplate` row to `STAFF_ONLY`. No pre-existing authored
    content exists today, but if it ever does the deploy needs an
    operator pass to flip live missions back to `OPEN`. Intentional —
    the audit step is the point of the gate.
  - **Adversarial review:** two passes during development (after the
    first three batches, then after B7). Both finished
    "MERGE WITH FIXES"; all minor findings addressed. No outstanding
    critical/major findings.

  ### Out of scope (deferred, queued behind this PR)

  - Full cross-app migration squash — attempted in this branch via
    reset-and-regenerate, broken by 13 apps' RunSQL/RunPython operations
    (materialized views, data backfills, table partitioning), rolled back
    cleanly. New memory `feedback_reset_regenerate_squash_unsafe` saved
    for next attempt; the safe path is `arx manage squashmigrations`
    (preserves RunSQL via `replaces=[...]`) or just leaving Phase B's
    9 incremental migrations as-is.
  - Phase C (predicate leaf-resolver expansion) and Phase D
    (Mission Studio backend API) are separate PRs off merged main.


src/world/missions/models.py
```diff
@@ -712,18 +935,63 @@ class MissionGiver(SharedMemoryModel):
...
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name="+",
+        help_text=(
+            "ENVIRONMENTAL_DETAIL kind: the examinable thing in the room "
+            "(real item or room detail) that kicks off the offer. Must be "
+            "null for other kinds. Should be an Item-typeclass ObjectDB "
+            "or a room-detail ObjectDB."
+        ),
```
TehomCD, line 973: Discriminator pattern only makes sense when the FKs can point to different tables. Why not just make it an ObjectDB and then we can determine the type of typeclass matching correctly based on the giver type?

TehomCD (dismissed):
nitpick about the MissionGiver discriminator (I don't really like two FKs to the same table where only one can be populated - seems nonsensical to me), but otherwise looks fine. I don't think it's blocking

TehomCD (approved):

PR #575, opened 2026-05-28
feat(missions): create-from-scratch UI + slug→PK sweep

ApostateCD:
## Summary

- **New Mission Studio create flow** — `/staff/missions/new` full-page form (`CreateMissionPage.tsx`) wired up from a `+ New Mission` button on `MissionBrowserPage`. All template-level fields collected, including categories via the new `CategoryMultiSelect`. Cooldown UX is amount + unit dropdown serialized to ISO 8601. On submit, navigates to `/staff/missions/<id>/canvas` so the author starts adding nodes.
- **Categories now editable end-to-end** — `MissionTemplateSerializer.categories` becomes DRF default `PrimaryKeyRelatedField` (read + write list of PKs); new `MissionCategoryViewSet` (`ReadOnlyModelViewSet` at `/api/missions/categories/`) feeds the picker; `EditCategoriesDialog` opens from a pencil button on the detail panel.
- **`MissionTemplate.slug` and `MissionGiver.slug` model fields removed** — PK becomes canonical. `MissionTemplateViewSet` / `MissionGiverViewSet` drop `lookup_field = "slug"`; custom action signatures `slug=` → `pk=`; `copy_template` service drops `new_slug` body field; `MissionGiver.name` gains `unique=True`. Frontend sweep: all api wrappers / query hooks / routes / URL params / page components updated. Two hand-written migrations (`0005_drop_slug_fields`, `0006_missiongiver_name_unique`) since `arx manage migrate` hangs on Evennia's setup wizard in the devcontainer.
- **Server-side auto-suffix on name collisions** for create + copy paths via a new `next_available_name(base, queryset)` service. PATCH renames still error 400 on collision (`validate_name`) — deliberate user choices deserve explicit feedback. New `validate()` proxy on the template serializer surfaces `MissionTemplate.clean()`'s level-band and `percent_replace` invariants as 400s instead of 500s.

## Spec & plan
- `docs/superpowers/specs/2026-05-27-missions-create-ui-and-slug-removal-design.md`

## Test plan
- [ ] `/staff/missions/` shows `+ New Mission` button; click navigates to `/staff/missions/new`
- [ ] Submit valid form → redirects to `/staff/missions/<id>/canvas` (new empty mission visible)
- [ ] Submit again with same name → toast `Saved as "X 2" — "X" was taken.`; URL has different PK
- [ ] Open detail panel for a template → pencil icon next to categories → dialog opens → toggle categories → save → panel updates
- [ ] Copy an existing template via `StaffActionsCard` — works without `new_slug` field; auto-suffixes name if collision
- [ ] Assign a mission to a character via staff-power — still works (URL uses pk)
- [ ] Open a node via `/staff/missions/<id>/nodes/<key>` — page loads (route param is now `:id`)
- [ ] Try PATCH-renaming a mission to a name already taken — expect 400 with inline `name` error (NOT auto-suffix, NOT 500)

## Tracked follow-ups (not blocking this PR)
- `just gen-api-types` should be rerun — `frontend/src/generated/api.d.ts` still has `slug` on MissionGiver; explicit interface overrides in `types.ts` carry the gap meanwhile.
- `validate_name()` duplicated on `MissionTemplateSerializer` + `MissionGiverSerializer` — minor DRY opportunity.
- `arx manage migrate` hangs on Evennia's superuser wizard in this devcontainer — friction-audit candidate.


.devcontainer/post-create.sh
```diff
@@ -14,8 +14,8 @@ eval "$(~/.local/bin/mise activate bash)"
 # author. ~/.gitconfig lives in the container's writable layer (not a named
 # volume), so this needs to re-run on every fresh container — cheap and
 # idempotent. Values mirror the host's bare-metal identity.
-git config --global user.name "Dave Brannigan"
-git config --global user.email "surly.mime@gmail.com"
+git config --global user.name "Dan Brannigan"
+git config --global user.email "apostatecd@yahoo.com"
```
TehomCD, line 18: lol can we have this be like some env var that gets copied in rather than us fighting to the death

TehomCD (approved):
looks fine to me but we should make the devcontainer github attribution be parameterized rather than hardcoded as either you or me

.devcontainer/post-create.sh
ApostateCD: follow up pr inc

PR #612, opened 2026-05-29
docs: keep systems reference docs mechanism-only (trim stale design narrative)

TehomCD:
## Summary

Docs hygiene. Reference docs under `docs/` should describe what the code does, not design rationale — design intent belongs in specs and tends to drift stale in reference material.

- `docs/roadmap/capabilities-and-challenges.md`: drop the editorial aside on the consequence pipeline; describe `select_weighted` + `build_outcome_display` (built from the full resolved pool) in plain mechanism terms.
- `docs/architecture/property-capability-action.md`: describe the absorbed `attempts`-app pattern as character-loss outcome filtering (mechanism) rather than editorializing.

No code changes; two doc lines.


ApostateCD (approved):
Yeah this is a sensible change.

PR #687, opened 2026-05-31
feat(magic): damage-type-scoped power modifiers (#655)

### LINKED ISSUE #655: Power: damage-type-scoped power modifiers (wire target_damage_type against TechniqueDamageProfile)
Follow-up carved out of #634/#635 (power as a derived term from scoped modifiers).

#634 ships **global + resonance-scoped** power modifiers in `_derive_power`. Damage-type scoping was **deliberately deferred** here, not forgotten.

## Why deferred
`ModifierTarget.target_damage_type` (FK → `conditions.DamageType`) already exists as a scope hook, and techniques DO carry damage types via `TechniqueDamageProfile` (`src/world/magic/models/techniques.py:692`, linking `technique` + `damage_type`). But wiring a damage-type power term means deciding how a single cast resolves its damage type (a technique can have a damage profile; the cast path in `use_technique` doesn't currently thread one into `_derive_power`). That's its own slice.

## Done when
- [ ] `_derive_power` (in `world/magic/services/techniques.py`) also sums power-scoped `ModifierTarget`s whose `target_damage_type` matches the cast's damage type.
- [ ] The cast path provides the damage type to match against (resolve how: read `TechniqueDamageProfile` for the technique; decide single-vs-multi damage type behavior).
- [ ] Reuses the same `get_modifier_total` summation + floor-at-0 as #634; power stays derived, never stored.
- [ ] Tests: damage-type-scoped power modifier raises landed effect on a matching-damage-type cast, and does NOT apply on a non-matching one; channeled intensity (anima/mishap/Soulfray) unaffected.

## Notes
- Mirror the resonance-scope helper added in #634; extend the same single-query + in-Python filter.
- Part of epic #633 (power vs intensity).

TehomCD:
## Summary

- Extends `_get_power_targets()` to `select_related("target_damage_type")` — single query, no extra round-trips
- Adds `technique_damage_type_ids` set in `_derive_power`, built from `technique.damage_profiles.all()` (untyped profiles excluded)
- AND-gate filter: a power `ModifierTarget` now must pass both resonance scope AND damage-type scope (null on either dimension = global)

Closes #655. Part of epic #633 (power vs intensity).

## Test Plan

- [ ] `DamageTypePowerDerivationTests` (7 new tests): matching profile raises power, non-matching skips, no profiles skips, untyped profile skips, any-of-multiple match, global target unaffected, channeled intensity unchanged
- [ ] All 28 tests in `test_power_derivation` + `test_audere_power_spike` pass (`just test-fast world.magic.tests.test_power_derivation world.magic.tests.test_audere_power_spike`)


ApostateCD (approved):
LGTM, damage types is a nice addition.

PR #852, opened 2026-06-08
fix(#745): clear three spread follow-ups (pagination, message XSS, seed cost)

ApostateCD:
Clears the three follow-ups that were noted in the Spread a Tale PR bodies (#781/#844/#845) but never filed — doing them now rather than leaving them as untracked debt in closed-PR history.

### 1. Pagination schema mismatch
`spreadable-deeds` and `spread-specializations` advertised `Paginated*List` responses in the OpenAPI schema (the viewset carries a `pagination_class`, so drf-spectacular infers pagination) but actually returned **bare lists**. Now they genuinely paginate via `paginate_queryset` / `get_paginated_response` — matching `deed-stories` (Phase 4) and the item-app `_paginated_response` precedent. The FE hooks read `.results`; the backend tests assert on `.results`. Schema and reality now agree.

### 2. `EvenniaMessage` XSS surface
The message renderer injected server HTML through `dangerouslySetInnerHTML` behind a literal *"sanitize it later if needed"* TODO. But chat messages carry **user-authored** text (poses, says, names) routed through Evennia's ANSI→HTML conversion, so the string is untrusted — a crafted pose was a stored-XSS vector. Now sanitized with **DOMPurify** (default config: strips `<script>`, event handlers, and dangerous tags while keeping the legitimate color `<span>`s). Adds the `dompurify` dependency. `ChatWindow`'s tests (which render `EvenniaMessage`) still pass under jsdom.

### 3. Lazy template-seed in the request path
`get_or_create_spread_a_tale_template()` + `ensure_spread_skills()` re-ran ~18 idempotent `get_or_create`s on **every** spread POST. The repo bans data migrations / seed commands pre-production, so they correctly stay seed-on-first-use — but a fast-path (one `.first()` fetch / one `.count()`) drops the steady state to a single query. Locked with an `assertNumQueries(1)` test.

### Testing
- PG parity: `world.scenes.tests.test_spread_endpoints` (12) + `world.societies.tests.test_spread_services` (7, incl. the new fast-path + idempotency tests) — green.
- FE `pnpm typecheck` + `pnpm build` + `ChatWindow` vitest — green. API types regenerated.

Refs #745 (closed umbrella) — pure cleanup, no new feature surface.

### DIFF
--- a/src/world/societies/spread_services.py
+++ b/src/world/societies/spread_services.py
@@ -73,8 +73,22 @@ def _ensure_skill(name: str):
 
 def ensure_spread_skills() -> None:
     """Idempotently ensure the spread skills (Performance, Persuasion) + their
-    form specializations exist."""
+    form specializations exist.
+
+    Called several times per spread (template build, modifiers, form list). The
+    repo bans data migrations / seed commands pre-production, so this is the
+    seed-on-first-use path — but the fast-path keeps the steady state at a single
+    count query instead of re-running the get_or_creates every request.
+    """
     from world.skills.models import Specialization  # noqa: PLC0415
+
+    form_names = [name for name, _, _ in SPREAD_FORMS]
+    already_seeded = Specialization.objects.filter(
+        name__in=form_names,
+        parent_skill__trait__name__in=[PERFORMANCE_SKILL_NAME, PERSUASION_SKILL_NAME],
+    ).count() == len(SPREAD_FORMS)
+    if already_seeded:
+        return
 
     skills_by_name = {
         PERFORMANCE_SKILL_NAME: _ensure_skill(PERFORMANCE_SKILL_NAME),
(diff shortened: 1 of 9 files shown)


src/world/societies/spread_services.py
```diff
@@ -157,6 +171,12 @@ def get_or_create_spread_a_tale_template():
     from world.checks.models import CheckCategory, CheckType, CheckTypeTrait  # noqa: PLC0415
     from world.traits.models import Trait, TraitCategory, TraitType  # noqa: PLC0415
 
+    # Fast-path: once seeded, the template (and its skills) exist, so a single
+    # fetch per spread suffices instead of replaying every get_or_create below.
+    existing = ActionTemplate.objects.filter(name="Spread a Tale").first()
+    if existing is not None:
```
TehomCD, line 179: Agree with that - we shouldn't be using String literals for constant values

TehomCD (approved):

PR #974, opened 2026-06-13
fix(magic): accept_thread_weaving_unlock bypasses the Mage Scar XP gate

TehomCD:
Closes #898

## Summary

Closes the Mage Scar XP-gate bypass. The gate (raise `AlterationGateError` on an OPEN `PendingAlteration`; `ProtagonismLockedError` when protagonism-locked) was enforced in `spend_xp_on_unlock` but not in the magic app's XP-spend services.

**Audit finding — the bypass was in TWO services, not one.** The issue scoped only `accept_thread_weaving_unlock` (which calls `xp_tracker.spend_xp`). A code-verified audit found `cross_thread_xp_lock` *also* spends XP — via `xp_tracker.total_spent += …; save()` — which is why a `.spend_xp(` grep missed it, and it bypassed the same gate. Gated both, so the 'all spend_xp sites enforce the gate or document why exempt' done-when actually holds.

**Changes:**
- Gate `accept_thread_weaving_unlock` (top of body) and `cross_thread_xp_lock` (after the idempotency early-return, before the spend — so already-unlocked/no-op calls aren't gated), mirroring `spend_xp_on_unlock` exactly.
- Surface the gate as HTTP 400 (standard gate message) in `AcceptTeachingOfferSerializer` and `CrossXPLockSerializer`, mirroring the established `resolve_audere_majora_offer` catch→`ValidationError` pattern.
- 8 TDD tests: gated path raises `AlterationGateError`, **no XP deducted, no unlock row created**, for both services; idempotent xp-lock bypasses the gate; teaching-offer accept endpoint returns 400 with the gate message.
- Marked the 'Known gap' note in `docs/roadmap/magic.md` resolved.

ty hook + full pre-commit clean on changed files.

## Follow-ups filed

(none)

## Notes

- Spec: reviewed & approved on #898 (in the issue body)
- Brainstorm/plan: skipped
- Sync-with-main: Rebased onto origin/main (clean, 1 commit).


TehomCD:
check: SonarCloud Code Analysis, signature: new-code duplication (gate body vs spend_xp_on_unlock), push: 40cea338

CI fix attempt 2: the residual duplication was my `enforce_advancement_gate` body still mirroring the inline gate in `spend_xp_on_unlock` (that pair survived attempt 1, which is why duplication only moved 10.2%→9.9%). Routed `spend_xp_on_unlock` through the shared helper — now a single gate implementation across all XP-spend sites. Behavior-identical (93 SQLite tests green; 2 unrelated integration tests need the Postgres tier for DISTINCT ON).

ApostateCD (approved):

Issue #1143, opened 2026-06-16
Epic: Investigation & Discovery — search, clues, research projects


ApostateCD:
# Epic: Investigation & Discovery — search · clues · research projects

The mystery/investigation core loop, built as one reusable spine. Players discover
hidden things (lore, missions, rescues, secrets) by acquiring **clues** and
**pursuing** them — sometimes instantly, sometimes through collaborative **research
projects**. Codex/lore is the primary consumer (the content mountain); rescue is the
cheapest first vertical; secrets/scandals come later.

## The model: a clue is a pointer, defined on three independent axes

1. **Target** — *what it points to*: a codex entry, a mission, a character secret, a
   scandal. **Invariant: a clue cannot exist without a target** (no red herrings, no
   empty clues). The target drives the **"you already know this"** flag the moment a
   clue would be surfaced.
2. **Acquisition** — *how you come to hold it*: meeting a precondition (high resonance
   entering an area, an old-soul tie to an item), a room **search**, or a random
   surfacing.
3. **Resolution** — *how holding it becomes having the target*: **automatic** (rescue
   clues, simple lore, "search again here") or a **research project** (the long haul).

Pipeline: **trigger → hold a clue → pursue it → gain the target.** Every stage is
pluggable, which is why rescue, lore, and a blackmail scandal are the *same machine*
with different target types and resolution lengths.

## The shared spine: capability-gated check → narrated spreadable deed

Discovery actions (a room **search**, a **research** contribution, a mission approach)
are all the **same primitive**: *an authored course-of-action a character attempts via
an **AP-priced**, **capability-gated** check that yields a player-narrated **deed**
(which spreads) plus a mechanical result.* Missions route that result through a graph;
**projects** fold it into progress. Projects must **reuse the mission deed/approach +
deed-spreading machinery**, not fork it.

This dissolves the projects `ContributionKind` AP-vs-CHECK confusion: a contribution
*is* the check; **AP is its cost**. "Ways to help" (haul materials = easy Strength;
oversee design = hard Intellect+architecture; dig archives = Intellect+Investigation)
are authorable **contribution methods** — different difficulties/capabilities, same
primitive. Resource pledges (pure money/item gifts) stay a separate, dumb shape.

## Capability gating rule (load-bearing — no sub-issue may reach past it)

Every option/method/approach gates on an **aggregated effective-capability value**
(e.g. effective Strength) and a difficulty — **never** on a specific gift or magic
[...]

PR #1161, opened 2026-06-17
feat(#1160): passive clue triggers — enter-room acquisition (slice A)

ApostateCD:
## #1160 slice A — passive clue triggers (enter-room acquisition)

The second acquisition surface beside search (#1154): a clue revealed on **entering a
room** because of *who you are / where you are* — no action, no roll, just a precondition.

### What this adds
- **`ClueTrigger`** (`room_profile` + `clue` + `eligibility_rule` predicate + `is_active`)
  — the passive counterpart to `RoomClue`. Kept as a **separate model** (considered
  unifying via a flag): same anchor data, but distinct acquisition semantics
  (search+roll vs passive+predicate), so one table never cross-contaminates the other.
- **`maybe_grant_clue_triggers`** — on entry, grants each active trigger the entrant is
  eligible for (empty rule = anyone) and hasn't held; AUTOMATIC clues resolve at once; the
  player is told via the clue's **authored description** (`send_narrative_message`).
- Wired into **`Character.at_post_move`** alongside the mission ROOM_TRIGGER dispatch +
  traps — same cheap room-bound query + best-effort (never breaks movement) rationale.

### Deferred (own slices on #1160)
Item-acquisition triggers (soul-tie / on-acquire), resonance/past-life predicate sources,
area-level (vs room) anchoring. Triggers + predicates are staff-editable data.

### Tests (`world.clues`, 40 green)
Eligible entrant granted; already-held not re-granted; ineligible / inactive / no-trigger
skipped; AUTOMATIC codex resolves on trigger; and an **on-move integration test** proving
the `at_post_move` wiring fires. Full-project `ty` clean.

Part of #1160. Part of #1143.


src/typeclasses/characters.py
```diff
@@ -353,6 +353,14 @@ def at_post_move(self, source_location, move_type="move", **kwargs):
 
                 maybe_emit_fame_reaction(self, self.location)
 
+            # Passive clue triggers (#1160): a room may reveal a clue to an
+            # eligible entrant — no search needed (who you are / where you are).
+            # Same cheap room-bound query + best-effort rationale as above.
+            with contextlib.suppress(Exception):
```
TehomCD: A little odd to skip *all* exceptions here. If we had a database query blow up due to a connection error or whatever, what would happen?

src/world/clues/services.py
```diff
@@ -222,3 +223,63 @@ def _roster_entry_for(character: ObjectDB) -> RosterEntry | None:
...
+        granted.append(clue)
+    return granted
+
+
+def _notify_clue_found(roster_entry: RosterEntry, clue: Clue) -> None:
+    """Tell the player a passive trigger revealed a clue (its authored description)."""
+    from world.narrative.constants import NarrativeCategory  # noqa: PLC0415
+    from world.narrative.services import send_narrative_message  # noqa: PLC0415
+
+    with contextlib.suppress(Exception):
```
TehomCD: It is strange to swallow *all* exceptions here. This could mask real errors

TehomCD (changes requested):
Swallowing/ignoring all exceptions isn't something you should normally do for a player-facing action

TehomCD (approved):
I still don't really get why we're handling errors at that stage, exactly, but at least now we're getting it logged. Random exception handling is sort of an antipattern since you're turning something into a no-op where it fails silently and players will just be confused. We'll get a log message at least, but it's better to have some surfaced error message to players while it's sent to something like Sentry. We don't have a specific error handling service yet so it's fine to use this for now but this is clearly not ideal

Issue #1162, opened 2026-06-17
Audit: convert silent suppress(Exception) best-effort hooks to log-and-continue


ApostateCD:
Follow-up to TehomCD's review on #1161. A best-effort hook that wraps a real operation in `contextlib.suppress(Exception)` (or `except Exception: pass`) keeps the action from breaking — but it also makes a genuine fault (a DB/connection error, a bug) **vanish silently**. We want "don't break the action" *and* "still know it broke". Convert these to **log-and-continue** (`try/except Exception: logger.exception(...)`).

## Sites to fix (game-side best-effort hooks)

- `typeclasses/characters.py` `at_post_move` — mission ROOM_TRIGGER dispatch (#729), traps (#1051), fame reactions (#881). *(The clue-trigger block in the same method was fixed in #1161.)*
- `typeclasses/mixins.py` — mission ENVIRONMENTAL_DETAIL dispatch on examine (#729).
- `world/missions/services/trigger_dispatch.py` `_announce` — mission grant notification.
- `world/stories/services/tables.py` — story offer notification.
- `world/narrative/services.py` — gemit `session.msg` loop (`except Exception: pass`).

## Out of scope (assessed, leaving)

- `cli/arx.py` (×3) — CLI-tool fallbacks, not game-facing hooks; intentional and noqa'd.
- `suppress` of a *specifically-expected* condition (e.g. `suppress(RosterEntry.DoesNotExist)`) — that's narrow and correct, not a silent-fail.

## Approach

Each site keeps continuing (best-effort), but logs the exception via a module `logger = logging.getLogger(__name__)`, so failures land in logs/monitoring instead of disappearing.

PR #1168, opened 2026-06-18
feat(#1164): bespoke error capture — SystemErrorReport + run_safely boundary

ApostateCD:
## #1164 — bespoke error capture (no-SaaS), per TehomCD's review

The destination TehomCD pointed at: best-effort hooks shouldn't *silently no-op* (confusing players) **or** break the action — they should **capture the fault for staff and tell the player**. No Sentry SaaS; we file our own reports.

### What this adds
- **`SystemErrorReport`** (`world/player_submissions`, next to `BugReport`) — the system files its own report: `traceback`, `label`, nullable `actor_persona`, a dedup `signature` + `occurrence_count`, `status`. Reviewed in the **same staff queue** as player bug reports (new `SubmissionCategory.SYSTEM_ERROR`). Admin registered.
- **`report_error(exc, *, label, actor)`** — structured log **+** a deduplicated report (a recurring error is one row with a count, not spam). Never raises.
- **`run_safely(label, fn, *, actor)`** — the boundary for optional/best-effort work. On failure it captures + (if `actor`) shows the player the verbatim line:
  > *An unexpected error has occurred. Staff has been notified, but please feel free to leave a bug report under the account menu.*
  …and continues — never breaking the player's primary action.
- **Migrated the player-facing hooks**: `Character.at_post_move` (mission ROOM_TRIGGER / traps / fame / clue triggers on enter) and the examine mission dispatch — replacing the ad-hoc `try/except: logger.exception` from #1162/#1163. Removed the now-unused module loggers there.

### Deferred (own slices, noted)
- Migrating the **notification-only** hooks (mission announce / story-offer / gemit / clue-found) to capture — they already log non-silently from the audit, so this is thin.
- A **staff frontend** view to triage `SystemErrorReport`s alongside BugReports.
- A **"file a GitHub issue from this report"** staff button.

### Tests (`world.player_submissions` + `typeclasses` + `world.clues`, 99 green)
`report_error` captures with traceback + dedups repeats into one counted row; `run_safely` returns on success, captures + returns None on failure, notifies the actor with the exact line, and captures-without-notifying when there's no actor. The move/examine integration paths still pass through the new boundary. Full-project `ty` clean.

Part of #1164.


TehomCD (approved):
I think this is okay for now. I don't really love it - sentry is a lot more elegant with its hooks into the process rather than the lambda wrappers we have here. Function decorators would probably be better for wrapping functions that needs to logs tracebacks if they error

PR #1224, opened 2026-06-19
Time-based cleanup of soft-deleted non-lore-critical items

TehomCD:
Closes #1025

## Summary

Adds time-based cleanup of soft-deleted, non-lore-critical `ItemInstance` rows (follow-up to #509).

- **`ItemInstance.is_lore_critical`** — never-purge predicate: `lore_value > 0` OR facets OR **transfer** provenance (`GIVEN`/`STOLEN`/`TRANSFERRED` only). `ACTIVATED`/`CONSUMED` are deliberately excluded — every consumed item has those, so counting them would make the sweep a silent no-op.
- **Shared `hard_delete_item_instance()` helper** — removes an instance's whole footprint (its ledger rows, then `game_object` CASCADE / direct delete) with **no dangling FKs**. Used by both the destruction-at-0 path (retiring #509's old nulled-`CONSUMED`-row behavior) and the new cleanup.
- **`purge_expired_soft_deleted_items()`** — daily `items.soft_delete_cleanup` cron task sweeps soft-deleted, past-grace, non-lore-critical rows. Grace = `ITEM_SOFT_DELETE_GRACE_DAYS` (.env, default 30). Per-item savepointed and skips PROTECT-referenced rows (Mantle / ProjectContribution) so one bad row can't wedge the batch.
- Soft-delete itself is unchanged: rows + FKs stay intact, just filtered; only the time-delayed purge removes them. No new migration. Docs updated in tandem. 426 items tests green.

## Follow-ups filed

(none)

## Notes

- Spec: reviewed & approved on #1025 (in the issue body)
- Brainstorm/plan: ran
- Sync-with-main: Rebased onto origin/main @ cfc3d5d8 (#1218) — no overlap with our files; clean.

### DIFF
--- a/docs/roadmap/items-equipment.md
+++ b/docs/roadmap/items-equipment.md
@@ -200,8 +200,44 @@ What landed:
   owner-or-staff gated, `ItemError` → HTTP 400; list/retrieve filter `.in_play()`.
 
 Deferred follow-ups: use-item *Action* (`actions/definitions/`) converging on `action.run()`
-alongside equip/unequip; time-based cleanup of soft-deleted, non-lore-critical instances;
-durability-on-use (#508) integration; quantity/stack consumption; frontend "use item" UI.
+alongside equip/unequip; durability-on-use (#508) integration; quantity/stack consumption;
+frontend "use item" UI.
+
+## Time-Based Cleanup of Soft-Deleted Items (DONE, #1025)
+
+**Branch:** `feature-1025-time-based-cleanup-of-soft-deleted-non-l`
+
+Closes the deferred cleanup follow-up from #509: non-lore-critical soft-deleted items are
+now automatically hard-purged after a configurable grace period.
+
+What landed:
+
+- **`is_lore_critical` predicate** on `ItemInstance` (`world/items/models.py`) — returns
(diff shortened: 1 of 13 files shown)


TehomCD:
✅ **CI green on all required checks** (every backend shard incl. `world.items`, `ty`, `pre-commit`, `api-types-drift`, CodeQL, SonarCloud). Auto-merge (squash) is armed — the merge queue will land this once a maintainer approves.

**Beyond the approved spec, two hardening passes were added during review:**
- **PROTECT-FK resilience** — the daily purge runs each delete in its own savepoint and excludes `Mantle` / `ProjectContribution`-referenced rows, so one PROTECT-referenced row can't roll back (and silently wedge) the whole batch.
- **In-world safety guard** — an otherwise-eligible row whose `game_object` still has a location (a half-undelete: re-homed into the world but `destroyed_at` still set) is **never** purged; it's logged at WARNING for staff to resolve. Prevents the cron deleting a game object out from under a live room.

**Conflict note:** rebased/merged cleanly after #1222 (use-item UI) landed — the only real overlap was the `use_item` docstring in `usage.py` (took #1222's) plus additive doc merges; `429→430` items tests green throughout.

Holding for review.

PR #1901, opened 2026-07-05
Thread crossing threshold requirements (#1885)

TehomCD:
## Summary

Adds an authored, per-thread-kind requirement gate at PathStage crossing levels (3, 6, 11, 16, 21) so the Rite of Imbuing (`spend_resonance_for_imbuing`) blocks advancement across a crossing until the character satisfies authored requirements (items, achievements, traits, etc.). This is the gating twin of the already-built resonance-based reward side (variant discoveries, signatures).

Closes #1885.

### New model: `ThreadCrossingThreshold`
- `world/magic/models/crossings.py` — authored catalog keyed on `(target_kind, level)`, with `clean()` validating the level is a crossing and `stage` matches `stage_for_level(level)`.
- Fail-open at runtime: no row for `(target_kind, level)` → no gate (mirrors Durance's `ClassLevelUnlock.DoesNotExist`).

### Polymorphic unlock target
- `AbstractClassLevelRequirement` → `AbstractUnlockRequirement` — `class_level_unlock` made nullable, `thread_crossing_threshold` FK added, CheckConstraint enforces exactly one is set.
- All 9 concrete requirement types inherit the polymorphic FK. `ItemRequirement` and `LegendRequirement` (which have their own `class Meta`) needed explicit constraint additions since their Meta overrides the abstract base's `constraints`.

### Service layer
- `_check_requirements(target, fk_name)` shared helper + `check_requirements_for_thread_crossing(character, threshold)` in `spends.py`.
- `is_crossing_level(level)` helper in `classes/services.py` — derives from `stage_for_level`.

### Imbuing loop gate
- Crossing check inserted after XP_LOCK, before cap check in `spend_resonance_for_imbuing`.
- `ThreadImbueResult` gains `"CROSSING_REQUIREMENT"` to `blocked_by` Literal + `blocked_requirement_messages: list[str]` field.
- `ImbueAction.execute` surfaces blocked messages with `success=True` (matching the XP_LOCK precedent — resonance was spent, partial advance happened).

### Admin + ADR + docs
- `ThreadCrossingThresholdAdmin` with `TraitRequirement` and `ItemRequirement` inlines in `progression/admin/unlocks_admin.py`.
- ADR-0090 documents the boundary choice, per-kind scoping, possession-only semantics, and ADR-0016 (shared base) vs ADR-0089 (sibling-per-domain) justification.
- `world/magic/CLAUDE.md` Thread System section updated.

## Key design decision

ADR-0089's "sibling per domain" was about **consumption semantics differing** — `RitualComponentRequirement` is consumed, `ItemRequirement` is possession-only. Thread crossings are possession-only too, so the distinguishing axis doesn't split Durance from thread crossings. ADR-0016's convergence rule applies — the same requirement family is reused via a polymorphic unlock target, not duplicated.

## Test coverage (47 tests, all passing)

- 20 model/service tests: constraints, `clean()` validation, `check_requirements_for_thread_crossing` (met/unmet/fail-open), related-name collision, Durance path unaffected
- 8 imbuing service tests: block-at-crossing, advance-when-met, fail-open-no-row, fail-open-no-requirements, partial-advance, item-not-consumed, level-6 crossing, per-kind scoping
- 3 action contract tests: crossing block message, pending-effect consumption, normal message
- 16 existing item-requirement regression tests (unchanged — proves no Durance breakage)

## Implementation note

Discovered that Django abstract base `Meta.constraints` are **not inherited** by concrete subclasses that declare their own `class Meta`. `ItemRequirement` and `LegendRequirement` both have their own `class Meta` (for `itemrequirement_exactly_one_mode` and `verbose_name` respectively), so the `exactly_one_unlock_target` constraint had to be explicitly added to each.


TehomCD:
All CI checks pass ✅ — 18/18 green (frontend-test skipped as expected). PR is enqueued for the merge queue; it will merge once a human approves the review.

PR #2895, opened 2026-08-02
Seed code mints Resonance/Affinity rows by name, and the content export ships them as authored content

TehomCD:
Closes #2890

## Summary

Closes the **export** half of ADR-0168. The seeding half was already closed and working:
every live seeder that touches a `CONTENT_MODELS` table already looks its rows up and
creates them only under `SEED_SAMPLE_CONTENT`, which defaults off.

The open half was that sampling is a **legitimate** path — it exists so a third party with
no content repo gets a bootable starter set — and once those rows exist, nothing
downstream told them apart from authored ones. So an export laundered invented names into
the corpus as lore.

That already happened. `magic/resonance.json` held twelve resonances, **none authored** —
eleven were sample rows, one came from a test-only path — while 22 of the 24 canonical
resonances were missing entirely and Praedari carried a canonically wrong affinity
(Primal; canon has it Abyssal, as Bene's opposite). A whole design round on #2736 was
built on that invented vocabulary. ArxII-lore#36 fixed the corpus; **without this PR the
next export undoes it.**

## The gate

`export_to_content_repo` treats a row whose natural key is not already in the corpus as an
**addition** and withholds it by default: written nowhere, reported in `result.withheld`.
Rows the corpus already knows still export, so edits round-trip exactly as before.
`allow_additions=True` is the authoring path and reports what it pushed in `result.added`.
Wired at both entry points — `--allow-additions` on `tools/export_content.py`, and a
checkbox on the admin export page, both listing what was withheld.

Refuses by default rather than warning: a warning printed after the write has already lost.

## Why this shape (ADR-0191)

The precise alternative was a **sample-provenance registry** — `authored_or_sample` has
exactly one `create` call serving 420 call sites, so recording `(model_label, natural_key)`
there is a one-line choke point, and it needs no confirmation step in the good case. It is
**forward-only**: it can never clean a database sampled before it landed, which is every
existing one, and its backfill would have no ground truth except "diff against the corpus"
— which is this design. That single argument decided it.

Also rejected: a per-model `is_sample` column (145 models, 28 apps); refusing to export
from any database ever sampled (a boolean that cannot be honestly unset); and exporting
only keys the corpus already has with no opt-in, which would stop staff authoring in admin
from ever reaching the corpus — `EXPORT_FILTERS`' own comment says that must keep working.

## The bug this nearly shipped with

Prose domains write one markdown file per entry, so "file absent" means "new entry" — and
on a virgin corpus that is true of **every** entry. My first draft keyed the gate per
entry and silently stopped every `StartingArea` and `CodexEntry` from exporting. It
surfaced only as an unrelated load-sequencing test losing its deferred-FK resolution;
I verified that test passes on clean main before treating it as mine. The gate keys on the
**domain** now, and `ExportAdditionGateProseDomainTests` names the case directly.

Two other deliberate bypasses: a model with no fixture file yet (genuinely new — every row
is a first export), and a model with no usable identity. A model whose rows are all
withheld writes nothing rather than `[]`, which would empty the corpus file outright.

## Scope notes

- **`NaturalKeyMixin.identity_fields()` is new** rather than a `# noqa: GETATTR_LITERAL`
  in the export module. The noqa ratchet's own message says to add the missing scaffolding
  instead of silencing it, and it was right: "can rows of this model be identified at all"
  is a question about the natural-key contract and belongs beside `natural_key()`.
- **Three stale content names fixed** — `ensure_cleanup_content` invented a Resonance named
  `"Hope"` (not one of the 24; in no seed cluster, so it never shipped, but it would have);
  `Resonance.name`'s `help_text` cited `'Sylva'`, also not one of the 24, shown to every
  staff member in admin; and `world/magic/CLAUDE.md` called Light/Sanctity/Radiance/
  Dissolution "canonical resonances" — none of the four is, and that line is plausibly how
  the confusion propagated in the first place.
- **`MagicContent`'s factory calls left alone, deliberately.** Its methods are called only
  from `integration_tests/` and `world/combat/tests/`, never from a seed cluster, and a
  test fixture builder *should* create its own rows. Gating it would mean 16 callers opting
  into sampling to get what factories already give. Its real problem is living under
  `world/seeds/` and reading as a seeder.

## Testing

`just test-fast core_management core world.areas` — 662/662 OK, plus 244/244 on
`core_management` after the rebase. New coverage: `ExportAdditionGateTests` (withhold /
allow / edit round-trip / never-empty-the-file) and `ExportAdditionGateProseDomainTests`.

`arx manage makemigrations --check` fails identically on clean main — pre-existing, not
from this branch.

## Follow-ups filed

(none)

## Notes

- Spec: reviewed & approved on #2890 (in the issue body)
- Brainstorm/plan: ran
- Sync-with-main: Rebased onto origin/main (3 commits replayed, no conflicts).


TehomCD:
check: api-types-drift, signature: stale src/schema.json + frontend/src/generated/api.d.ts, push: d7dc75c9c

`Resonance.name`'s `help_text` feeds the DRF serializer description, so it lands in the
committed OpenAPI schema and the generated frontend types. Correcting the stale `'Sylva'`
example was a one-line docs-accuracy fix with a generated-artifact dependency I didn't
account for.

Regenerated with `just gen-api-types`; the diff is exactly that one description, nothing else.

Note on the previous run: the `backend` aggregate reported `cancelled`, not failed — the
shards were superseded by the SonarCloud fix push mid-run.

PR #3022, opened 2026-08-05
Content loads never overwrite credited rows; sample seeding requires an empty content universe

TehomCD:
Closes #3017

## Summary

Content loads never overwrite credited rows: any differing field on a row whose written_by is set freezes the whole row into a conflicts report (CLI + admin flash), and the only override is a per-row, typed-confirmation, digest-guarded delete-then-reload page in admin - no bulk resolve exists. The freeze covers all three write paths that touch CreditedContent rows (fixture/markdown upsert, grid TriggerDefinition refresh, weather emits - the last also gaining credit-name resolution that fixes a real crash on raw natural-key credit refs, plus a credited-row skip in the crafting-materials seeder). Sample seeding now hard-refuses to run once any CONTENT_MODELS table has rows within a press. ADR-0201 records the rulings, the accepted costs (no sync bookkeeping, scan isolation, per-press gate scope), and the one deliberate bypass: admin Import Data remains an undocumented-no-longer, superuser-only disaster-recovery overwrite path - flagged for a future guard if Tehom wants one.

## Follow-ups filed

(none)

## Notes

- Spec: reviewed & approved on #3017 (in the issue body)
- Brainstorm/plan: ran
- Sync-with-main: Rebased onto origin/main cleanly (branch not yet pushed); no conflicts, no migration files on this branch.

### DIFF
--- a/docs/adr/0201-credited-rows-are-frozen-against-loads.md
+++ b/docs/adr/0201-credited-rows-are-frozen-against-loads.md
@@ -43,8 +43,11 @@ name is dropped and logged, never the row.
 **Trade-offs / accepted cost:**
 - No sync bookkeeping means the freeze cannot distinguish "someone hand-edited this row in the
   live DB and never exported it" from "the repo's fixture genuinely moved past what's in the DB" -
-  both look identical (a credited row, a differing incoming value) and both route through the same
-  resolution page. Ruled acceptable by Tehom, 2026-08-05, on #3017.
+  both look identical (a credited row, a differing incoming value), and each path reports it
+  through its own channel rather than one shared resolution page: the fixture pipeline surfaces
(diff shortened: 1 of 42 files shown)


TehomCD:
CI green on all 6 checks; auto-merge (squash) armed. Built via subagent-driven execution with per-task adversarial review: the review loop caught and fixed two coverage holes the plan missed (grid TriggerDefinition refresh branch, crafting-materials seeder stomping credited ItemTemplates), one real crash bug (raw natural-key credit refs in the weather seed), and added the stale-digest guard on the resolve flow. One deliberate bypass is documented, not guarded: admin Import Data (disaster-recovery overwrite, superuser-only, own dry-run preview) - see ADR-0201's trade-offs; say the word if you want the freeze extended there in a follow-up. Human approval is the only remaining gate.

PR #3207, opened 2026-08-17
Registration must survive a mail failure, and staff must be able to copy a verification link

### LINKED ISSUE #3193: Registration must survive a mail failure, and staff must be able to copy a verification link
## Problem

A failed verification email breaks registration completely.

Allauth sends a verification email during signup. On 2026-08-16 the send raised
`TimeoutError: [Errno 110] Connection timed out`, and the exception propagated
out of `/api/auth/browser/v1/auth/signup` as an HTTP 500. Django returns an HTML
error page for a 500, so the React client failed to parse it and displayed
`Unexpected token '<', "<!DOCTYPE "... is not valid JSON`.

The account row and the invite redemption commit before the send runs. The user
therefore received a real account, a spent invite, and a failure message. A
second attempt then fails with "username taken", which reads as a different bug.

PR #3192 fixes the transport that caused this specific outage. It does not fix
the fragility: any mail provider failure still turns a successful registration
into an apparent failure.

Allauth also sends a verification email at login
(`send_verification_email_at_login`), so the same failure blocks an existing
user from signing in.

## Proposed direction

Two independent changes. Both need a spec before implementation.

### 1. A mail failure must not fail the request

Registration succeeds or fails on its own terms. Treat mail delivery as a
separate, best-effort step.

- Catch send failures at the registration and login boundaries.
- Return a success response that reports the account state and states that the
  verification mail could not be sent.
- Record the failure where staff can see it, so a silent mail outage does not
  stay invisible.
- The client must show a useful message and never a raw parser error. It must
  also tolerate a non-JSON body rather than calling `res.json()` on it
  unconditionally.

Open question for the spec: whether the user should be told to contact staff, or
whether the interface should offer a retry.

### 2. Staff can copy a verification link

Email should not be the only route to a verified account.

Registration is invite-only and staff-gated, so staff already decide who joins.
A staff-copyable verification link matches the trust model of the invite link
that staff already paste by hand. It also removes a third-party dependency from
the critical path.

- Add a staff surface that produces the verification link for an account.
- Reuse the existing verification route. The frontend already has
  `EmailVerifyPage` at `/verify-email/:key`, and allauth can generate the key on
  demand.
- Mirror the invite-link surface from #3181 rather than inventing a new pattern.

Open question for the spec: whether the link is single-use and time-bounded, and
whether generating one is recorded for audit.

## Why this is not folded into another PR

This changes registration and login behavior and adds a staff surface. It needs
its own spec and its own review.

## Evidence

Production log, 2026-08-16:

```
[ERROR] Internal Server Error: /api/auth/browser/v1/auth/signup
  File "allauth/account/internal/flows/email_verification.py", line 294,
    in send_verification_email_at_login
TimeoutError: [Errno 110] Connection timed out
```

TehomCD:
Closes #3193

## Summary

Two changes from the 2026-08-16 outage. (1) Mail is best-effort at the allauth boundary: ArxAccountAdapter.send_mail catches provider failures, logs them, and records them in the new AccountMailFailure model (read-only admin) - a dead mail provider no longer turns a committed signup into an HTML 500, and no longer blocks login via send_verification_email_at_login. The React client also stops feeding non-JSON bodies to res.json() at the login/signup boundaries. (2) Staff can copy a verification link: POST /api/staff/verification-link/ generates allauth's HMAC confirmation key on demand for an unverified address (time-bounded, audit-logged), surfaced as a Copy Verification Link form on the staff invites page - mirroring the #3181 invite Copy Link fallback. Adds migration 0156 (AccountMailFailure). BROAD_EXCEPT ratchet 113->114, justified in the commit message. Spec skipped per Tehom (ops-hardening from a production outage).

## Follow-ups filed

(none)

## Notes

- Spec: reviewed & approved on #3193 (in the issue body)
- Brainstorm/plan: skipped
- Sync-with-main: Branch cut from current main; no sync needed.


TehomCD:
check: api-types-drift, signature: generated api.d.ts stale after new /api/staff/verification-link/ endpoint, push: f9397ebb7

